PR #88, opened 2014-12-25
Fix various windows build issues

e9925248:


lilrc:
I think you will have to redo the PortAudio-fix commit since the build system has changed since you opened this PR. But otherwise the PortAudio-fix looks correct.

adiknoth:
If you still care about this PR, please update to current master and then ping the thread, so that we can merge shortly after.

PR #92, opened 2015-01-13
support [unescaped] double-quotes in jackdrc

x42:
allow autostart of jackd on OSX where device-names can
contain spaces.


sletz:
I trust you...

PR #110, opened 2015-04-03
Add a --no-samplerate build option

lilrc:
The build option --no-samplerate can be used to not build against libsamplerate even if it is available. This is necessary to avoid an automagic dependency which is bad, see [1]. Also pkg-config is used to find libsamplerate instead of the header check. Note that the default behaviour is not changed; if --no-samplerate is not given (default) waf will detect samplerate and build against it conditionally, as usual.

If this change is welcome I will also fix the automagic dependencies on celt, opus sndfile and readline. All these changes are required to fulfill quality assurance in Gentoo, so it would be great if you merged this request to ease my undertaking as maintainer of jack2 in the Pro-Audio Gentoo Overlay [2].

[1] https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies
[2] http://proaudio.tuxfamily.org/wiki/index.php?title=Main_Page

Thanks in advance,
Karl

### DIFF
--- a/common/wscript
+++ b/common/wscript
@@ -6,11 +6,6 @@ import re
 import os
 
 def configure(conf):
-    conf.check_cc(header_name='samplerate.h', define_name="HAVE_SAMPLERATE")
-   
-    if conf.is_defined('HAVE_SAMPLERATE'):
-        conf.env['LIB_SAMPLERATE'] = ['samplerate']
-
     conf.env['BUILD_NETLIB'] = conf.is_defined('HAVE_SAMPLERATE')
     conf.env['BUILD_ADAPTER'] = conf.is_defined('HAVE_SAMPLERATE')
 
--- a/wscript
+++ b/wscript
@@ -80,6 +80,7 @@ def options(opt):
     opt.add_option('--clients', default=64, type="int", dest="clients", help='Maximum number of JACK clients')
     opt.add_option('--ports-per-application', default=768, type="int", dest="application_ports", help='Maximum number of ports per application')
     opt.add_option('--debug', action='store_true', default=False, dest='debug', help='Build debuggable binaries')
+    opt.add_option('--no-samplerate', action='store_true', default=False, dest='no_samplerate', help='Do not build against libsamplerate if available')
     opt.add_option('--firewire', action='store_true', default=False, help='Enable FireWire driver (FFADO)')
     opt.add_option('--freebob', action='store_true', default=False, help='Enable FreeBob driver')
     opt.add_option('--alsa', action='store_true', default=False, help='Enable ALSA driver')
@@ -141,7 +142,6 @@ def configure(conf):
     conf.env.append_unique('CXXFLAGS', '-Wall')
     conf.env.append_unique('CFLAGS', '-Wall')
 
-    conf.sub_config('common')
     if conf.env['IS_LINUX']:
         conf.sub_config('linux')
         if Options.options.alsa and not conf.env['BUILD_DRIVER_ALSA']:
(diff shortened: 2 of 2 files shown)


mseeber:
I don't know if this works for all corner-cases, but this looks good to me.
:+1:

lilrc:
Do you want to merge this separately and merge the other fixes later, or should I add fixes for the other automagic dependencies to this pull request before merging?

mseeber:
I can not merge pull requests for jack2, I am only expressing what I think about this PR. From my point of view, this kind of improvement is worth the effort, but I am not a maintainer of jack2.

Someone able to merge will probably pick this up in a few days, I guess. If you are unsure if these changes are appreciated, you can also ask on the JACK mailing list or on IRC.

nedko:
I agree with the intent of this pull request but I have three small concerns:

1. the commit message it is not properly formatted. the summary line should be shorter. the message should contain info about why conf.sub_config('common') line is moved and about why LIB_SAMPLERATE is not explicitly defined anymore.
2. the historic precedent with optional dependencies (1ddb418fabc84dee8a3d07ebf45efbabc4378d90) is to have them disabled by default. I personally don't like this behavior but it is probably better to have consistent approach and eventually have it changed in a different commit. I think it makes sense to have three-state approach for optional dependencies: auto (enable if available, default), no/false/disable and yes/true/enable.
3. it may be good idea to have the single-check part of changeset in different commit

I won't merge until @sletz says his opinion on point 2.
Then @lilrc should adjust, to match my concerns 1 and 2 (unless @sletz is ok with the inconsistent behavior). I'll merge even if 3 is not fixed.

sletz:
1) please keep consistent behavior as much as possible

2) Nedko, you better know than me what to do with those waf scripts in the first place�

St�phane 

Le 5 avr. 2015 � 23:14, Nedko Arnaudov <notifications@github.com> a �crit :

lilrc:
The problem is that it is on one hand nifty to enable shallow automagic dependencies by default so that as much as possible is built-in when users manually compile, but that also requires --no-* flags since waf does not (by default) supply "two sided flags", see further down. On the other hand if positive flags (like --alsa, --firewire, ...) should always be used the options must be disabled by default creating some inconveniences for users who might have to enable a lot of flags to get just a few tools.

As I see it there are three (reasonable) ways to fix the inconsistency issue.

1) Let the --no-samplerate be. When other automagic deps are fixed later on there will be more --no-something options, making the approach more consistent since there will be a distinction between (one) backends that are a disabled by default but can be enabled with flags and (two) smaller dependencies (samplerate, sndfile, readline, ...) that are enabled by default (if possible), but can be disabled if requested.

2) Use a --samplerate flag and disable samplerate by default. In this case all other automagic deps will eventually need to be disabled by default (celt, opus, readline, sndfile).

3) Create a new kind of option that works approximately like this (this is still a draft!): --with-foo=yes|no|auto. If --with-foo=yes is given foo is included. If --with-foo=no is given foo is not included. By default --with-foo=auto, which means the build system autodetects whether foo is available and then builds with foo if it is.

I can settle with any of these, but I would first like to discuss them a little.

1) A quick and small solution and does not bother users or impose disabled-by-defaults, but with a small loss of consistency.

2) Also a quick and small fix, but bothers users with disabled-by-defaults, but is very consistent.

3) This will require more waf-scripting and to make it consistent the current options --alsa, --firewire, ... would have to be migrated to this kind of options first (or else it wouldn't be very consistent, right?). This would ofcourse remove current options and replace them with other ones, possibly confusing people, but it gives great consistency together with sane defaults; anything can be enabled by default, even the backends (--alsa, --firewire, ...), at the same time as no automagic dependencies are introduced. In the end I think this kind of options is the most user friendly as the build system will detect and build without the need to specify options such as --iio or --portaudio. The problem is that waf by default does not supply this kind of options, so they have to be wrapped in some way (around a string, presumably) so that everything can be handled neatly and consistently.

In case of 1) I will fix the issues @nedko pointed out. In case of 2) I will just change option and the default together with fixing the other pointed out issues. In case of 3) I will throw away this commit and work on the interface/framework that will be used for the options and then migrate the current options and then presenting a pull request. After that then I will fix the automagic dependency issues (samplerate, sndfile, readline and so on...).

Please, tell me what you think! I'd love to hear your thoughts!

nedko:
I rule out 1) as both me and sletz prefer consistency. I think having sane defaults and ability to disable unwanted dependencies (3) is much more important for the project than the samplerate detection issue itself. So please, go with 3. I prefer --foo=yes|no|auto over --with-foo=yes|no|auto because it is shorter, because this is what is currently used and because "--foo" could be treated as "--foo=yes".

lilrc:
I will be back with another pull request.

PR #125, opened 2015-06-02
[jack_bufsize] Fix off by 10 error

mseeber:
Since help and error messages suggest, that 8192 samples is a valid
buffer size, this fixes the value in the check to 8192 instead of 8182.


mseeber:
This problem does not apply to jack1 as jack1 uses a different upper limit.

PR #135, opened 2015-07-19
Avoid conflicts when encoding/decoding sync packet

jmendeth:
This doesn't affect the code behaviour, just a little aesthetics change.

In `EncodeSyncPacket` instead of writing to `fTxData` directly, each encoding step now writes to `sync_data`. This initially points to the start of `fTxData`, but each encoding step moves the pointer forward, so next steps will write after that area.

Also, `fPacketSize` is not set at `EncodeSyncPacket` (which would be expected since the packet size depends on how much space the encoding steps consume) but at `SyncSend`. This has been fixed, `fPacketSize` is now set dynamically in `EncodeSyncPacket`.

After this change, one can add encoding steps to the method without needing to update the other steps and the packet size.

### DIFF
--- a/common/JackNetInterface.cpp
+++ b/common/JackNetInterface.cpp
@@ -464,7 +464,6 @@ namespace Jack
         fTxHeader.fSubCycle = 0;
         fTxHeader.fDataType = 's';
         fTxHeader.fIsLastPckt = (fParams.fSendMidiChannels == 0 && fParams.fSendAudioChannels == 0) ? 1 : 0;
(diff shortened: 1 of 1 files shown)


jmendeth:
Any insight? The patch is simple, it just changes code like this:

~~~ c
data[0] = 0x56;
data[1] = 0x10;
data[2] = 0x20;
data[3] = 0x84;
~~~

into:

~~~ c
int i = 0;
data[i++] = 0x56;
data[i++] = 0x10;
data[i++] = 0x20;
data[i++] = 0x84;
~~~

See how in the second block you can just comment out one line and remove that byte from the output.

jmendeth:
Closing, I see you touched that part of the file, but somehow Github sees no conflicts...

PR #141, opened 2015-09-06
Account for IP+UDP header in MTU

jmendeth:
`PACKET_AVAILABLE_SIZE` now also subtracts `UDP_HEADER_SIZE`, to prevent NetJack from sending too long packets.

### DIFF
--- a/common/JackNetTool.h
+++ b/common/JackNetTool.h
@@ -49,8 +49,7 @@ using namespace std;
 #define UDP_HEADER_SIZE 64   // 40 bytes for IP header in IPV6, 20 in IPV4, 8 for UDP, so take 64
 #define HEADER_SIZE (sizeof(packet_header_t))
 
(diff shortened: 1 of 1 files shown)


sletz:
Le 6 sept. 2015 � 10:10, Xavier Mendez <notifications@github.com> a �crit :

> PACKET_AVAILABLE_SIZE now also subtracts UDP_HEADER_SIZE, to prevent NetJack from sending too long packets.

What does *that* mean?

What are you retying to to in the first place ??

St�phane

jmendeth:
When we calculate the available room for data in a NetJack packet, we currently take the MTU (1500 by default) and subtract the size of the NetJack header. That's wrong, because the IP and UDP headers also take space in these 1500 bytes, we should subtract their length as well.

This is a `TODO:` in the code actually, if you look at the diff you'll notice I'm only doing what it says.

jmendeth:
This is a one-liner, I'm just resolving the TODO added by @sletz on [54ada6](https://github.com/jackaudio/jack2/commit/54ada6e5450c71fea6f2eef98b6661720915efb4#diff-a2a2cc5841f69b8c83e54045cea0118c).

PR #142, opened 2015-09-10
Correct typos in JackNetTool.cpp

jmendeth:
Missing parentheses in `JackNetTool.cpp`, see patch.

### DIFF
--- a/common/JackNetTool.cpp
+++ b/common/JackNetTool.cpp
@@ -778,8 +778,8 @@ namespace Jack
                 memset(fCompressedBuffer[port_index], 0, fCompressedMaxSizeByte * sizeof(char));
             }
 
-            int res1 = (fNPorts * fCompressedMaxSizeByte + CDO) % PACKET_AVAILABLE_SIZE(params);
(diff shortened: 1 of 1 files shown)


joshdekock:
@sletz Is it actually meant to be `((fNPorts * fCompressedMaxSizeByte) + CDO)` or will this actually fix a typo?

jmendeth:
I'm confident this was a typo. `CDO` are the 2 bytes for the length. Since every port has these 2 bytes before the data, it only makes sense to do `fNPorts * (fCompressedMaxSizeByte + CDO)`.

That's why, if you look at the rest of the code, `CDO` is only used in `(fCompressedMaxSizeByte + CDO)`.

sletz:
Robin Gareus wrote the code.

Le 11 mars 2016 � 13:13, Xavier Mendez <notifications@github.com> a �crit :

adiknoth:
@x42 Hey Robin!

If you still remember anything from that code, it should be easy for you to click the merge button (or reject if you know the original code is correct).

adiknoth:
I assume @jmendeth has looked more carefully into this than anyone else by now. Merging this.

PR #145, opened 2015-09-27
Fix building with celt and opus in non standard prefix

karllinden:
This makes it possible to compile JACK with celt or opus even if they are not available in the standard system prefix (usually /usr).

This fixes celt and opus includes that leaked into the JackNetDriver (which is built without celt and opus) by moving celt and opus code into separate files. Also the opus header handling had to be changed since the opus/ prefix is already included in the includepath from the pkg-config opus file.

Please review this. Especially the code move, so I have not missed anything.

Thanks.


sletz:
1) Why this ? (seems strange�)

-#include <opus/opus.h>
-#include <opus/opus_custom.h>
+#include <opus.h>
+#include <opus_custom.h>

2) this will break OSX and Windows builds which have to be adapted before merge can be done.

St�phane 

Le 27 sept. 2015 � 08:50, Karl Lind�n <notifications@github.com> a �crit :

lilrc:
The "opus/" part is already included in the include path that is acquired from the opus pkg-config file.

```
   1 # Opus codec reference implementation pkg-config file
   2 
   3 prefix=@prefix@
   4 exec_prefix=@exec_prefix@
   5 libdir=@libdir@
   6 includedir=@includedir@
   7 
   8 Name: Opus
   9 Description: Opus IETF audio codec (@PC_BUILD@ build)
  10 URL: http://opus-codec.org/
  11 Version: @VERSION@
  12 Requires:
  13 Conflicts:
  14 Libs: -L${libdir} -lopus
  15 Libs.private: @LIBM@
  16 Cflags: -I${includedir}/opus
```

Above is opus.pc.in [1] from the opus git repo. Note the last line where the include path is set to the includedir (ususally /usr/include) with the /opus dir added. Say opus is installed into /home/user/opus. If the includes were not changed the compiler would not find opus/opus.h since the pkg-config file would induce the -I/home/user/opus/include/opus into C(XX)FLAGS and there is no such header file /home/user/opus/include/opus/opus/opus.h. It should be /home/user/opus/include/opus/opus.h.

Are Windows and Mac not using pkg-config?

From [2].
> pkg-config works on multiple platforms: Linux and other UNIX-like operating systems, Mac OS X and Windows. It does not require anything but a reasonably well working C compiler and a C library, but can use an installed glib if that is present. (A copy of recent glib2 is shipped together with pkg-config versions since 0.27, and this is sufficient for pkg-config to compile and work properly.) 

[1] https://git.xiph.org/?p=opus.git;a=blob;f=opus.pc.in;h=e15c6fed215e32a21d77f8cc24e9fd63db52f534;hb=HEAD
[2] https://wiki.freedesktop.org/www/Software/pkg-config/

sletz:
This " Cflags: -I${includedir}/opus" seems wrong to me, first time I see that… AFAICS the correct way is to have  Cflags: -I${included} and let the developer use <opus/opus.h> kind of naming.

lilrc:
Well, we do not choose what is in opus' pkg-config files, do we? ;) It is the opus developers choice. This is the include path we are given to use the opus library and we should use it, IMHO.

If we do not use it we end up with headers not found, or worse, compiling against the wrong opus headers. Consider a system where there are two opus versions installed (maybe the user wants to test how different versions behave, or is triaging a bug, or something...). Suppose they are in two prefixes `PrefixA` and `PrefixB`. If we do not use the `-I` flag given to us by the opus pkg-config and blindly `#include <opus/opus.h>` we may end up with including `PrefixA/include/opus/opus.h` when the desired is `PrefixB/include/opus/opus.h`, or vice versa.

Of course this might seem like a far-fetched example, but still, we get this corner case (and also the case where I discovered this issues) for free.

lilrc:
Also I think we should follow the include style of other open source projects utilising opus and omitting the leading "opus/" part, such as VLC [1], opusfile (!) [2], MPD [3], FFmpeg [4][5][6], just to name few.

Also in the opus demo [7] the "opus/" is omitted

[1] https://git.videolan.org/?p=vlc.git;a=blob;f=modules/codec/opus.c;h=1e93cd4da1e43fd60bc67e77da09c32980c8438d;hb=HEAD
[2] https://git.xiph.org/?p=opusfile.git;a=blob;f=include/opusfile.h;h=a31b6c47f6d8e70bade3887de4a05838c9bf2422;hb=HEAD
[3] http://git.musicpd.org/cgit/master/mpd.git/tree/src/decoder/plugins/OpusDecoderPlugin.cxx
[4] https://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavcodec/libopus.c;h=16395c73dfc7a181c156afbf08abc40f442e03c9;hb=HEAD
[5] https://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavcodec/libopusdec.c;h=4cbd14a38a8d0ca445f1e6a00608af4947f72423;hb=HEAD
[6] https://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavcodec/libopusenc.c;h=8ac02f95205bdfa37bfa719e73c0cf8ed5f68f72;hb=HEAD
[7] https://git.xiph.org/?p=opus.git;a=blob;f=src/opus_demo.c

sletz:
Still� I would be interested to know why the Opus developers choose that way...

Le 28 sept. 2015 � 17:16, Karl Lind�n <notifications@github.com> a �crit :

sletz:
I understand for sure, my point is "did the Opus developer choose that way by design ? or is is a stupid error of them ?"


Le 28 sept. 2015 � 17:40, Karl Lind�n <notifications@github.com> a �crit :

lilrc:
It makes sense since if there would be an opus-1 someday that would be (somewhat) API incompatible consumer programs would just have to bump their pkg-config check to "opus-1" and there would be no need to change any `#includes`. Also as you see this forces consumer programs to be consistent with the `#includes` and consistency is nice.

Also consider the case where some user decides to install arbitrary stuff into a prefix, which we may call `prefix`. If the user installs "some crazy libraries with headers that will break stuff" to prefix and also decides to put an innocent version of opus into it then without the opus developers choice of `-I`-path we would end up scanning `prefix` for all includes which would mean we could end up including "headers that break stuff". Far-fetched, maybe, but possible.

lilrc:
Is there anything blocking this, or can we proceed?

sletz:
It breaks OSX and Windows build, i would like to fix that before merging

lilrc:
Why doesn't Windows and OSX use the pkg-config file of opus? It is the most standard way to compile against libraries, and if a library has a pkg-config file it should clearly be used (since it might contain extra dependencies, flags, paths, etc.) that cannot be detected using simpler checks.

sletz:
??

Because OSX uses Xcode project and Windows Code blocks. Do you want to fix those 10 years old projects ? or redo them?

Le 8 oct. 2015 � 15:24, Karl Lind�n <notifications@github.com> a �crit :

sletz:
OSX El capitan support is more urgent for me right now, and I have little time to give to JACK project.

lilrc:
I think waf can generate an xcode project and I would not be surprised if it can generate code blocks too, but waf is designed to be cross-platform so using waf directly should work too. Atleast the build system contains a lot of iffery for windows and osx. It would be nice to have the same build system for all platforms, really...

If someone can try it I can make waf generate an xcode. It should be simple, I hope. Refer to [1].

[1] https://stackoverflow.com/questions/22189764/generate-xcode-project-with-waf

lilrc:
Can we please get somewhere with this? Wouldn't it be better if we could all use the waf build system. It is system independent, so it runs on all operating systems.

If someone wants an xcode project, we can start experimenting with exporting an xcode project using waf. However, I need help on testing and verifying this, since I do not run OSX.

Is anyone using the Code Blocks thing? There is code to handle Windows builds in the waf build system, so there seems to have been work gone into making it work with Windows.

What are the reasons to keep these other build systems, anyway? To me it seems to be an unnecessary maintenance burden. Besides, I think it is silly to have three build systems for one project.

sletz:
Le 28 nov. 2015 � 08:54, Karl Lind�n <notifications@github.com> a �crit :

> Can we please get somewhere with this? Wouldn't it be better if we could all use the waf build system. It is system independent, so it runs on all operating systems.
> 
> If someone wants an xcode project, we can start experimenting with exporting an xcode project using waf. However, I need help on testing and verifying this, since I do not run OSX.
> 
> Is anyone using the Code Blocks thing? There is code to handle Windows builds in the waf build system, so there seems to have been work gone into making it work with Windows
> 
> What are the reasons to keep these other build systems, anyway? To me it seems to be an unnecessary maintenance burden. Besides, I think it is silly to have three build systems for one project.
> 

The reason? Nobody has time or energy to change the current situation.

St�phane

joshdekock:
<details> 
  <summary>Click to show patch</summary>
```diff
diff --git a/waflib/extras/xcode.py b/waflib/extras/xcode.py
new file mode 100644
index 0000000..4a97577
--- /dev/null
+++ b/waflib/extras/xcode.py
@@ -0,0 +1,312 @@
+#! /usr/bin/env python
+# encoding: utf-8
+# XCode 3/XCode 4 generator for Waf
+# Nicolas Mercier 2011
+
+"""
+Usage:
+
+def options(opt):
+	opt.load('xcode')
+
+$ waf configure xcode
+"""
+
+# TODO: support iOS projects
+
+from waflib import Context, TaskGen, Build, Utils
+import os, sys, random, time
+
+HEADERS_GLOB = '**/(*.h|*.hpp|*.H|*.inl)'
+
+MAP_EXT = {
+	'.h' :  "sourcecode.c.h",
+
+	'.hh':  "sourcecode.cpp.h",
+	'.inl': "sourcecode.cpp.h",
+	'.hpp': "sourcecode.cpp.h",
+
+	'.c':   "sourcecode.c.c",
+
+	'.m':   "sourcecode.c.objc",
+
+	'.mm':  "sourcecode.cpp.objcpp",
+
+	'.cc':  "sourcecode.cpp.cpp",
+
+	'.cpp': "sourcecode.cpp.cpp",
+	'.C':   "sourcecode.cpp.cpp",
+	'.cxx': "sourcecode.cpp.cpp",
+	'.c++': "sourcecode.cpp.cpp",
+
+	'.l':   "sourcecode.lex", # luthor
+	'.ll':  "sourcecode.lex",
+
+	'.y':   "sourcecode.yacc",
+	'.yy':  "sourcecode.yacc",
+
+	'.plist': "text.plist.xml",
+	".nib":   "wrapper.nib",
+	".xib":   "text.xib",
+}
+
+
+part1 = 0
+part2 = 10000
+part3 = 0
+id = 562000999
+def newid():
+	global id
+	id = id + 1
+	return "%04X%04X%04X%012d" % (0, 10000, 0, id)
+
+class XCodeNode:
+	def __init__(self):
+		self._id = newid()
+
+	def tostring(self, value):
+		if isinstance(value, dict):
+			result = "{\n"
+			for k,v in value.items():
+				result = result + "\t\t\t%s = %s;\n" % (k, self.tostring(v))
+			result = result + "\t\t}"
+			return result
+		elif isinstance(value, str):
+			return "\"%s\"" % value
+		elif isinstance(value, list):
+			result = "(\n"
+			for i in value:
+				result = result + "\t\t\t%s,\n" % self.tostring(i)
+			result = result + "\t\t)"
+			return result
+		elif isinstance(value, XCodeNode):
+			return value._id
+		else:
+			return str(value)
+
+	def write_recursive(self, value, file):
+		if isinstance(value, dict):
+			for k,v in value.items():
+				self.write_recursive(v, file)
+		elif isinstance(value, list):
+			for i in value:
+				self.write_recursive(i, file)
+		elif isinstance(value, XCodeNode):
+			value.write(file)
+
+	def write(self, file):
+		for attribute,value in self.__dict__.items():
+			if attribute[0] != '_':
+				self.write_recursive(value, file)
+
+		w = file.write
+		w("\t%s = {\n" % self._id)
+		w("\t\tisa = %s;\n" % self.__class__.__name__)
+		for attribute,value in self.__dict__.items():
+			if attribute[0] != '_':
+				w("\t\t%s = %s;\n" % (attribute, self.tostring(value)))
+		w("\t};\n\n")
+
+
+
+# Configurations
+class XCBuildConfiguration(XCodeNode):
+	def __init__(self, name, settings = {}, env=None):
+		XCodeNode.__init__(self)
+		self.baseConfigurationReference = ""
+		self.buildSettings = settings
+		self.name = name
+		if env and env.ARCH:
+			settings['ARCHS'] = " ".join(env.ARCH)
+
+
+class XCConfigurationList(XCodeNode):
+	def __init__(self, settings):
+		XCodeNode.__init__(self)
+		self.buildConfigurations = settings
+		self.defaultConfigurationIsVisible = 0
+		self.defaultConfigurationName = settings and settings[0].name or ""
+
+# Group/Files
+class PBXFileReference(XCodeNode):
+	def __init__(self, name, path, filetype = '', sourcetree = "SOURCE_ROOT"):
+		XCodeNode.__init__(self)
+		self.fileEncoding = 4
+		if not filetype:
+			_, ext = os.path.splitext(name)
+			filetype = MAP_EXT.get(ext, 'text')
+		self.lastKnownFileType = filetype
+		self.name = name
+		self.path = path
+		self.sourceTree = sourcetree
+
+class PBXGroup(XCodeNode):
+	def __init__(self, name, sourcetree = "<group>"):
+		XCodeNode.__init__(self)
+		self.children = []
+		self.name = name
+		self.sourceTree = sourcetree
+
+	def add(self, root, sources):
+		folders = {}
+		def folder(n):
+			if not n.is_child_of(root):
+				return self
+			try:
+				return folders[n]
+			except KeyError:
+				f = PBXGroup(n.name)
+				p = folder(n.parent)
+				folders[n] = f
+				p.children.append(f)
+				return f
+		for s in sources:
+			f = folder(s.parent)
+			source = PBXFileReference(s.name, s.abspath())
+			f.children.append(source)
+
+
+# Targets
+class PBXLegacyTarget(XCodeNode):
+	def __init__(self, action, target=''):
+		XCodeNode.__init__(self)
+		self.buildConfigurationList = XCConfigurationList([XCBuildConfiguration('waf', {})])
+		if not target:
+			self.buildArgumentsString = "%s %s" % (sys.argv[0], action)
+		else:
+			self.buildArgumentsString = "%s %s --targets=%s" % (sys.argv[0], action, target)
+		self.buildPhases = []
+		self.buildToolPath = sys.executable
+		self.buildWorkingDirectory = ""
+		self.dependencies = []
+		self.name = target or action
+		self.productName = target or action
+		self.passBuildSettingsInEnvironment = 0
+
+class PBXShellScriptBuildPhase(XCodeNode):
+	def __init__(self, action, target):
+		XCodeNode.__init__(self)
+		self.buildActionMask = 2147483647
+		self.files = []
+		self.inputPaths = []
+		self.outputPaths = []
+		self.runOnlyForDeploymentPostProcessing = 0
+		self.shellPath = "/bin/sh"
+		self.shellScript = "%s %s %s --targets=%s" % (sys.executable, sys.argv[0], action, target)
+
+class PBXNativeTarget(XCodeNode):
+	def __init__(self, action, target, node, env):
+		XCodeNode.__init__(self)
+		conf = XCBuildConfiguration('waf', {'PRODUCT_NAME':target, 'CONFIGURATION_BUILD_DIR':node.parent.abspath()}, env)
+		self.buildConfigurationList = XCConfigurationList([conf])
+		self.buildPhases = [PBXShellScriptBuildPhase(action, target)]
+		self.buildRules = []
+		self.dependencies = []
+		self.name = target
+		self.productName = target
+		self.productType = "com.apple.product-type.application"
+		self.productReference = PBXFileReference(target, node.abspath(), 'wrapper.application', 'BUILT_PRODUCTS_DIR')
+
+# Root project object
+class PBXProject(XCodeNode):
+	def __init__(self, name, version):
+		XCodeNode.__init__(self)
+		self.buildConfigurationList = XCConfigurationList([XCBuildConfiguration('waf', {})])
+		self.compatibilityVersion = version[0]
+		self.hasScannedForEncodings = 1;
+		self.mainGroup = PBXGroup(name)
+		self.projectRoot = ""
+		self.projectDirPath = ""
+		self.targets = []
+		self._objectVersion = version[1]
+		self._output = PBXGroup('out')
+		self.mainGroup.children.append(self._output)
+
+	def write(self, file):
+		w = file.write
+		w("// !$*UTF8*$!\n")
+		w("{\n")
+		w("\tarchiveVersion = 1;\n")
+		w("\tclasses = {\n")
+		w("\t};\n")
+		w("\tobjectVersion = %d;\n" % self._objectVersion)
+		w("\tobjects = {\n\n")
+
+		XCodeNode.write(self, file)
+
+		w("\t};\n")
+		w("\trootObject = %s;\n" % self._id)
+		w("}\n")
+
+	def add_task_gen(self, tg):
+		if not getattr(tg, 'mac_app', False):
+			self.targets.append(PBXLegacyTarget('build', tg.name))
+		else:
+			target = PBXNativeTarget('build', tg.name, tg.link_task.outputs[0].change_ext('.app'), tg.env)
+			self.targets.append(target)
+			self._output.children.append(target.productReference)
+
+class xcode(Build.BuildContext):
+	cmd = 'xcode'
+	fun = 'build'
+
+	def collect_source(self, tg):
+		source_files = tg.to_nodes(getattr(tg, 'source', []))
+		plist_files = tg.to_nodes(getattr(tg, 'mac_plist', []))
+		resource_files = [tg.path.find_node(i) for i in Utils.to_list(getattr(tg, 'mac_resources', []))]
+		include_dirs = Utils.to_list(getattr(tg, 'includes', [])) + Utils.to_list(getattr(tg, 'export_dirs', []))
+		include_files = []
+		for x in include_dirs:
+			if not isinstance(x, str):
+				include_files.append(x)
+				continue
+			d = tg.path.find_node(x)
+			if d:
+				lst = [y for y in d.ant_glob(HEADERS_GLOB, flat=False)]
+				include_files.extend(lst)
+
+		# remove duplicates
+		source = list(set(source_files + plist_files + resource_files + include_files))
+		source.sort(key=lambda x: x.abspath())
+		return source
+
+	def execute(self):
+		"""
+		Entry point
+		"""
+		self.restore()
+		if not self.all_envs:
+			self.load_envs()
+		self.recurse([self.run_dir])
+
+		appname = getattr(Context.g_module, Context.APPNAME, os.path.basename(self.srcnode.abspath()))
+		p = PBXProject(appname, ('Xcode 3.2', 46))
+
+		for g in self.groups:
+			for tg in g:
+				if not isinstance(tg, TaskGen.task_gen):
+					continue
+
+				tg.post()
+
+				features = Utils.to_list(getattr(tg, 'features', ''))
+
+				group = PBXGroup(tg.name)
+				group.add(tg.path, self.collect_source(tg))
+				p.mainGroup.children.append(group)
+
+				if 'cprogram' or 'cxxprogram' in features:
+					p.add_task_gen(tg)
+
+
+		# targets that don't produce the executable but that you might want to run
+		p.targets.append(PBXLegacyTarget('configure'))
+		p.targets.append(PBXLegacyTarget('dist'))
+		p.targets.append(PBXLegacyTarget('install'))
+		p.targets.append(PBXLegacyTarget('check'))
+		node = self.srcnode.make_node('%s.xcodeproj' % appname)
+		node.mkdir()
+		node = node.make_node('project.pbxproj')
+		p.write(open(node.abspath(), 'w'))
+
+
diff --git a/wscript b/wscript
index 4925238..fa995cc 100644
--- a/wscript
+++ b/wscript
@@ -387,7 +387,8 @@ def options(opt):
     # options provided by the modules
     opt.load('compiler_cxx')
     opt.load('compiler_c')
+
+    opt.load('xcode', help="Generate an Xcode 3/4 project file")
 
     # install directories
     opt.add_option('--htmldir', type='string', default=None, help="HTML documentation directory [Default: <prefix>/share/jack-audio-connection-kit/reference/html/")
```
</details>

Here's a patch for adding Xcode functionality. To apply: `patch -p1 < xcode.patch` assuming you are in the root dir of jack, and the patch is called `xcode.patch`
Then it works like: `python waf configure xcode`, and a Xcode project will generated. It will only support Xcode 3/4 project though, this is due to a limitation of this version of waf. If you upgrade waf to the latest, it will have native support for Xcode 6 projects

lilrc:
@joshdekock This  looks very nice. I really appreciate it. However, I cannot test it, but I assume you have tested it, which you have, right?

Are there any old xcode files lying around which should be deleted when this patch is applied? I guess we can finally replace the old xcode build system with this one. Am I correct? (I do not know very much about xcode, honestly.)

Just so we are clear, the latest version is 1.8.17. I will upgrade waf to the latest version and enable xcode6 when this one gets going.  When I do so I will file a pull request. Can I ping you Josh, so you can test it?

@sletz Can we please ignore codeblocks and go on? Is there anyone caring about it? If the waf build system does not work on windows I expect people to report bugs. I am happy to fix any issues with waf, if I just get the reports and some help testing stuff.

sletz:
Le 11 d�c. 2015 � 21:24, Karl Lind�n <notifications@github.com> a �crit :

> @joshdekock This looks very nice. I really appreciate it. However, I cannot test it, but I assume you have tested it, which you have, right?
> 
> Are there any old xcode files lying around which should be deleted when this patch is applied? I guess we can finally replace the old xcode build system with this one. Am I correct? (I do not know very much about xcode, honestly.)
> 
> Just so we are clear, the latest version is 1.8.17. I will upgrade waf to the latest version and enable xcode6 when this one gets going. When I do so I will file a pull request. Can I ping you Josh, so you can test it?
> 
> @sletz Can we please ignore codeblocks and go on? Is there anyone caring about it?

I do since this is the way the Windows version is compiled.

> If the waf build system does not work on windows I expect people to report bugs. I am happy to fix any issues with waf, if I just get the reports and some help testing stuff.
> 

Sorry a patch will not be accepted until it allows to do what what done with the previous system on OSX and Windows.

St�phane

joshdekock:
@lilrc 
> However, I cannot test it, but I assume you have tested it, which you have, right?
Of course.
> Can I ping you?
Yeah, sure. 
> Are there any old xcode files lying around which should be deleted when this patch is applied?
I didn't even know there was another way to generate an Xcode project. Granted, I only started looking at the code today.
Also, I'm in #jack on Freenode as ethe a lot, if that helps?

adiknoth:
Do you still care about this PR? If so, please rebase and ping the thread. Otherwise just close.

karllinden:
I will create a new PR when I want the fixes to be included.

PR #157, opened 2015-11-10
fix missing void parameter in prototypes

andrewrk:


mseeber:
@pauldavisthefirst This would also apply to jack_get_time() in the jack1 headers, if it gets accepted?https://github.com/jackaudio/headers/blob/master/jack.h#L1274

andrewrk:
https://github.com/jackaudio/headers/pull/6

PR #182, opened 2016-01-01
Fix various string initialization bugs

milasudril:
Valgrind complains about unitialized writes. This change should fix that. I also removed some dangerous `strcpy`. If fUUID always should be zero terminated there are some more fixes to do.

### DIFF
--- a/common/JackRequest.h
+++ b/common/JackRequest.h
@@ -110,14 +110,14 @@ struct JackRequest
     }
 
     virtual int Write(detail::JackChannelTransactionInterface* trans) { return -1; }
-    
+
     virtual int Write(detail::JackChannelTransactionInterface* trans, int size)
     {
         fSize = size;
         CheckRes(trans->Write(&fType, sizeof(RequestType)));
         return trans->Write(&fSize, sizeof(int));
     }
-    
+
     virtual int Size() { return 0; }
 
 };
@@ -168,7 +168,9 @@ struct JackClientCheckRequest : public JackRequest
     JackClientCheckRequest(const char* name, int protocol, int options, int uuid, int open = false)
         : JackRequest(JackRequest::kClientCheck), fProtocol(protocol), fOptions(options), fUUID(uuid), fOpen(open)
     {
-        snprintf(fName, sizeof(fName), "%s", name);
+        memset(fName, 0, sizeof(fName));
+        strncpy(fName, name, JACK_CLIENT_NAME_SIZE + 1);
+        fName[JACK_CLIENT_NAME_SIZE]=0;
     }
 
     int Read(detail::JackChannelTransactionInterface* trans)
@@ -190,7 +192,7 @@ struct JackClientCheckRequest : public JackRequest
         CheckRes(trans->Write(&fUUID, sizeof(int)));
         return trans->Write(&fOpen, sizeof(int));
     }
-    
+
     int Size() { return sizeof(fName) + 4 * sizeof(int); }
 
 };
@@ -210,7 +212,9 @@ struct JackClientCheckResult : public JackResult
     JackClientCheckResult(int32_t result, const char* name, int status)
             : JackResult(result), fStatus(status)
     {
-        snprintf(fName, sizeof(fName), "%s", name);
+        memset(fName, 0, sizeof(fName));
+        strncpy(fName, name, JACK_CLIENT_NAME_SIZE + 1);
+        fName[JACK_CLIENT_NAME_SIZE]=0;
     }
 
     int Read(detail::JackChannelTransactionInterface* trans)
@@ -246,7 +250,9 @@ struct JackClientOpenRequest : public JackRequest
     {}
     JackClientOpenRequest(const char* name, int pid, int uuid): JackRequest(JackRequest::kClientOpen)
     {
-        snprintf(fName, sizeof(fName), "%s", name);
+        memset(fName, 0, sizeof(fName));
+        strncpy(fName, name, JACK_CLIENT_NAME_SIZE + 1);
+        fName[JACK_CLIENT_NAME_SIZE]=0;
         fPID = pid;
         fUUID = uuid;
     }
@@ -268,7 +274,7 @@ struct JackClientOpenRequest : public JackRequest
     }
 
     int Size() { return 2 * sizeof(int) + sizeof(fName); }
-    
+
 };
 
 /*!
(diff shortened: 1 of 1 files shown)


MrMormon:
@milasudril

mseeber:
Doesn't snprintf already ensure that the string is null terminated? Why replace one line that is completely fine with 3 lines?
The fact that valgrind can not understand that the buffer does not need to be initialized doesn't mean that the code is not correct.
What's the rationale besides "looks better with valgrind" ? Maybe I am missing something.

The strcpy changes on the char* arguments look a bit safer on a first look though.

sletz:
+1 => we have to change only if a real problem exists….

falkTX:
there are errors in valgrind when using jack, so it's not possible to get a good output with it.
I think we should fix them whenever possible, but this pull request is overkill.

milasudril:
@mseeber It depends if we rely on strict C99 conformance. If we are using MSVCRT, then `snprintf` follows C89 and then it does not write `\0` if it does not fit.

milasudril:
Related: #208

mseeber:
@milasudril Ah, I see, MSDN states that snprintf is conform to C99 with VS2015 but good to know. If jack2 is built on older compilers this is relevant then. If jack2 needs to compile on thes C99 non conformant


Valgrind uses suppression files for false positives, fixing a shortcomings of valgrind should not be objective of jack2, I personally think.
Chapter 2.5 of the core manual mentions how to do this, maybe the more clean option would be to provide suppression files for these false positives? Maybe these are already included with valgrind/glibcand just need to be turned on?

Alternatively, one could just initialize (zero) the mentioned buffers once before using them. This should be also less heavy than zeroing them on every function call.

gazoo74:
Well... it is not **only** a matter of setting or not a trailing *NUL* character ; it is **also** a matter of writing a buffer that is not fully initialized.

Jack is writing the **whole** buffer to the socket instead of writing **only** the null-terminated string till the *NUL* character. The space between the trailing *NUL* character and the end of the buffer is left uninitialized, that why valgrind is complaining. Thus this is not a false positive.

IMO, the cleanest solution is to write **only** the null-terminated string and not the whole buffer.
```
CheckRes(trans->Write(&fName, strlen(fName)));
```
But this solution breaks the read function as well!

So a simple solution consist in writing the length first and then the *null-terminated* string
```
int Write(...)
{
  //...
  size_t len = strlen(fName);
  CheckRes(&len, sizeof(len));
  CheckRes(trans->Write(&fName, len));
  //...
  return 0;
}
```
```
int Read(...)
{
  //...
  size_t len;
  CheckRes(&len, sizeof(len)));
  CheckRes(trans->Read(&fName, sizeof(fName)));
  // ...
  return 0;
```

gazoo74:
This is simply a suggestion.

BTW, I wonder why it is the object (data) which knows how to serialize itself to/using the *JackChannelTransactionInterface*? Should not be the *opposite*?

I mean:
```
// Somewhere else in the code, but not in JackRequest objects...
void sendRequest(...)
{
  // ...
  JackRequest *req = new JackClientCheckRequest(...);
  JackChannelTransactionInterface *trans = new JackChannelTransactionImpl(...);
  trans->Read(req, sizeof(JackClientCheckRequest));
  // ...
}
```

I will save a lots of amount of code in JackRequest.h: no needs to implement Read/Write methods for all JackRequests...

Maybe I am wrong.

milasudril:
Regarding snprintf, there are some places in the code where this function can be replaced by strncpy, but observe that "No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow)."

Maybe the cleanest solution is to a pure memcpy instead, and force the last byte to be NUL afterwards. This will avoid the extra memset, assuming that the source buffer is filled according to this model.

gazoo74:
Hi guys, any news?

milasudril:
I realized that memcpy will be problematic for the caller, since then "A string constant" needs to be padded up to the required length. Another good approach is to create a write wrapper:

    void write_string(char* buffer,N,const char* source)
        {
        memset(buffer,0,N);
        strncpy(buffer,souce,N);
        buffer[N-1]=0;
        }

I should have done that, but I did not see that every call was wrong until I read through the whole file. I think that C99 functions should be avoided on Windows, since MinGW usually relies on MSVCRT.DLL, which is C89 only. Maybe `#define __USE_MINGW_ANSI_STDIO 1`

gazoo74:
Maintainers, @mseeber, any news ???

mseeber:
Sorry I can not make a decision since:
* there are also considerations to be made about target platforms. (see the MSVC thing)
* Also currently, I do not  have enough free time to read into the source at the respective places to make an informed decision.

The only recommendation I would give is, to split the non functional changes, e.g. white spaces and formatting into a separate commit.

adiknoth:
I've merged @gazoo74's pull request. If there's anything left to do in this PR, please update the PR and ping the thread.

7890:
I don't want to ponder on this, but a small test shows that using memset makes a significant difference vs. only strncpy + end termination.
Please run this. Could be that the test is wrong, but my numbers show around 30 % difference in runtime for a common case.
[null_minimal_2.c.gz](https://github.com/jackaudio/jack2/files/2798730/null_minimal_2.c.gz)
If these methods will be called millions of times, it can make a difference.

adiknoth:
> my numbers show around 30 % difference in runtime for a common case.
> If these methods will be called millions of times, it can make a difference.

These methods are called three times per hour. No need to micro-optimise for a few nanoseconds.

To be specific: they are called whenever a jack port is (re)named. That happens at creation time for each audio channel and then rarely ever again: if you rename your track in Ardour from "Drums" to "Drumkit", then you indirectly call this method. How often does that happen? As I said, three times per hour. ;)

7890:
To me it seems that it's touching the whole buffer twice, which is around 50 % too much :) I'm still speculating though. This is a recurring pattern.

7890:
When always terminating the array manually, we could then also use strncpy for every case. (No more using strncpy).
```
strncpy(arr, foo, sizeof(arr));
arr[sizeof(arr)-1]=0;
```
foo can be char *foo or char foo[n]. No overflow possible in arr, bytes padded, always terminated, touching the whole buffer only once. It seems safe, are there any other insights? I'm not about to change it (yet hehe), but for good understanding.

7890:
This pull request can be closed (again :) since the initialization issues for JackRequest.h are fixed.
The rest is code style questions in the widest sense. They need to be addressed at another place (if needed at all). It's not wrong to keep an eye on "clean" strings in all of the code. Note to self: static variables differ for initialization behavior. Inspiration:
```
    char* JackTools::UserDir()
    {
        static char user_dir[JACK_PATH_MAX + 1] = "";

        /* format the path name on the first call */
        if (user_dir[0] == '\0') {
            if (getenv ("JACK_PROMISCUOUS_SERVER")) {
                snprintf(user_dir, sizeof(user_dir), "%s/jack", jack_tmpdir);
            } else {
                snprintf(user_dir, sizeof(user_dir), "%s/jack-%d", jack_tmpdir, GetUID());
            }
        }

        return user_dir;
    }
```

PR #195, opened 2016-03-11
Fix Netjack crashes due to broken alignment

jmendeth:
Part of 02f74a659 touched the `_session_params` struct, breaking the alignment and causing both ends (driver and manager) to crash at negotiation. ~~This reverts the relevant change.~~

### DIFF
--- a/common/JackNetTool.h
+++ b/common/JackNetTool.h
@@ -38,7 +38,7 @@ using namespace std;
 #endif
 #endif
 
-#define NETWORK_PROTOCOL 8
+#define NETWORK_PROTOCOL 9
 
 #define NET_SYNCHING      0
 #define SYNC_PACKET_ERROR -2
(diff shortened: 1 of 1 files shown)


sletz:
This could be *much* better two fix the change, rather than reverting it...

jmendeth:
So, the PR now adds +8 as suggested in the mailing lists.
I'm not sure about bumping the protocol version, but I think it's appropiate since we're modifying the parameters structure.

adiknoth:
We should add a little more verbose explanation in the commit that adds +8. Feel free to copy&paste from my e-mail or link to <http://comments.gmane.org/gmane.comp.audio.jackit/29516>, hoping that this link will never go away. ;)

Because there will be somebody questioning the +8 in three years down the road, and then we won't remember.

Otherwise good and ready for merging.

joshdekock:
Maybe add the note in the next release notes? As 1.9.10 is fairly old, and with a netjack protocol change it might be an idea to create another release.

jmendeth:
> We should add a little more verbose explanation in the commit that adds +8. Feel free to copy&paste from my e-mail or link to http://comments.gmane.org/gmane.comp.audio.jackit/29516, hoping that this link will never go away. ;)

I've linked to the thread.

PR #206, opened 2016-05-16
Make backtrace support depends on execinfo.h existence

tSed:
In some C-libraries (like uclibc), backtrace support is optional, so the
execinfo.h may not exist.

This change adds the check for execinfo.h header and conditionaly enable
backtrace support.

This issue has been triggered by Buildroot farms:
  http://autobuild.buildroot.org/results/391/391e71a988250ea66ec4dbee6f60fdce9eaf2766/build-end.log

Signed-off-by: Samuel Martin s.martin49@gmail.com

### DIFF
--- a/wscript
+++ b/wscript
@@ -483,6 +483,12 @@ def configure(conf):
         defines=['_GNU_SOURCE'],
         mandatory=False)
 
+    # Check for backtrace support
+    conf.check(
+        header_name='execinfo.h',
+        define_name='HAVE_EXECINFO_H',
+        mandatory=False)
+
     conf.recurse('common')
     if Options.options.dbus:
         conf.recurse('dbus')
--- a/dbus/sigsegv.c
+++ b/dbus/sigsegv.c
@@ -27,7 +27,9 @@
 #include <stdio.h>
 #include <signal.h>
 #include <dlfcn.h>
-#include <execinfo.h>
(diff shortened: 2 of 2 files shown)


wscript
```diff
@@ -483,6 +483,12 @@ def configure(conf):
         defines=['_GNU_SOURCE'],
         mandatory=False)
 
+    # Check for backtrace support
+    conf.check_cc(
```
lilrc: Do not use `check_cc` use `check` instead. See https://github.com/waf-project/waf/blob/master/waflib/Tools/c_config.py#L762

lilrc:
See comment on the `wscript` file. Otherwise it looks good.

wscript
```diff
@@ -483,6 +483,13 @@ def configure(conf):
         defines=['_GNU_SOURCE'],
         mandatory=False)
 
+    # Check for backtrace support
+    conf.check(
+        compiler='c',
```
lilrc: Why compiler='c'? Waf should be able to guess that. See https://github.com/waf-project/waf/blob/master/waflib/Tools/c_config.py#L472

lilrc:
I'm glad you changed from `check_cc` to `check`, but see my comment.

wscript
tSed: Well, I'm not familiar with waf...

tSed:
ping?

PR #208, opened 2016-05-19
Valgrind

gazoo74:
Dear Maintainers,

I was running jackd using valgrind when I notice several complains. I fixed them into this patch-set (except fd48059f9bd4f78d0c350fdc30cd37363c09365f).

Feel free do modify the patch set if you disagree with my fixes. You probably need to review the default value I set into the commit 9e5d1909d156ff730cd3d0cb10cdbe489f7004ed.

I hope my patches will be useful.

Regards,
Gaël

### DIFF
--- a/common/JackRequest.h
+++ b/common/JackRequest.h
@@ -163,8 +163,10 @@ struct JackClientCheckRequest : public JackRequest
     int fUUID;
     int fOpen;
 
-    JackClientCheckRequest()
-    {}
+    JackClientCheckRequest() : fProtocol(0), fOptions(0), fUUID(0), fOpen(0)
+    {
+        memset(fName, 0, sizeof(fName));
+    }
     JackClientCheckRequest(const char* name, int protocol, int options, int uuid, int open = false)
         : JackRequest(JackRequest::kClientCheck), fProtocol(protocol), fOptions(options), fUUID(uuid), fOpen(open)
     {
@@ -207,7 +209,9 @@ struct JackClientCheckResult : public JackResult
     int fStatus;
 
     JackClientCheckResult(): JackResult(), fStatus(0)
-    {}
+    {
+        memset(fName, 0, sizeof(fName));
+    }
     JackClientCheckResult(int32_t result, const char* name, int status)
             : JackResult(result), fStatus(status)
     {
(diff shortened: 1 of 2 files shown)


milasudril:
Related: #182

gazoo74:
Yes I saw your PR afterward.

Thanks

gazoo74:
BTW, my patch-set contains also:
- a91969e12ddc5eb8b6e0aac4458c5de52c0f1a86: a fix which triggers an unexpected process callback,
- fd48059f9bd4f78d0c350fdc30cd37363c09365f: a duplicate instruction and
- 9e5d1909d156ff730cd3d0cb10cdbe489f7004ed: that fix extra valgrind complains.

gazoo74:
ping ;)

milasudril:
More on memory initialization:

I can confirm that `strncpy` is the safest bet. The man page says:

> If  the length of src is less than n, strncpy() writes additional null bytes to dest to ensure that a total of n bytes are written.

This is also true for non-c99 MSVCRT (tested on old Windows 98, with MinGW 3.4.5). The recipe for a safe string copy is simply

    strncpy(dest,source,length);
    buffer[length-1]='\0';

If snprintf is ever needed (the string has to be formatted somehow) use `memset`, `snprintf`, and

    buffer[length-1]='\0';

This should always work.

gazoo74:
Any news maintainers?

At least, the change a91969e fix a bug that triggers an unexpected process callback.

common/JackRequest.h
```diff
@@ -1310,16 +1355,18 @@ struct JackSessionNotifyRequest : public JackRequest
...
     {
+        memset(fPath, 0, sizeof(fPath));
+        memset(fDst, 0, sizeof(fDst));
         snprintf(fPath, sizeof(fPath), "%s", path);
+        fPath[JACK_MESSAGE_SIZE] = 0;
         if (dst) {
             snprintf(fDst, sizeof(fDst), "%s", dst);
-        } else {
-            fDst[0] = '\0';
+            fDst[JACK_CLIENT_NAME_SIZE] = 0;
```
adiknoth: Is this semantically correct?

Old: terminating the char array at the first position
New: terminating the char array at the last (second to last?) position (some stuff is declared as SIZE+1)

gazoo74: Yes, this is semantically correct.
- I dropped the old _} else {_ condition at line 1321,
- and the old instruction _fDst[0] = '\0';_ at line 1322 is now performed by the new instruction _memset(fDst, 0, sizeof(fDst));_ at line 1364.

Both instructions:
1. _fPath[JACK_MESSAGE_SIZE] = 0;_
2. _fDst[JACK_CLIENT_NAME_SIZE] = 0;_
make sure strings are NUL terminated. In fact, those paranoid instructions can be dropped.

PR #209, opened 2016-05-19
wscript: replace deprecated use of check_cc by check

tSed:
Using check_cc function is not recommended by the waf project [1], the
check function should be use instead.

So, replace all occurrences of:
  check_cc(...)
by:
  check(compiler='c', ...)

[1] https://github.com/waf-project/waf/blob/master/waflib/Tools/c_config.py#L762

Signed-off-by: Samuel Martin <s.martin49@gmail.com>


lilrc:
This is already fixed. See 4e8e7ec.

PR #219, opened 2016-07-13
Use linux futex as JackSynchro

### LINKED ISSUE #183: 64bit jackd2 and 32bit clients
Is it possible to use 64bit jackd2 with 32bit clients like Renoise, vsthost or Wine? I'm on Ubuntu and if I use that configuration I got no sound and midi.

falkTX:
The current use of posix semaphores in the jack2 code is wrong.
See https://sourceware.org/bugzilla/show_bug.cgi?id=17980

> The type sem_t has never been compatible between 32-bit and 64-bit processes, neither size nor alignment match.

The old "working" behavior was never officially supported.

This patch adds a new class that uses a linux futex as JackSynchro, and makes it the default for linux builds.
I made it so that internal clients use the 'private' futex flag, this is specially useful in ARM systems where context switches are very expensive.

Fixes #183.

PS: I've been testing this change myself for more than a month, and it is now the default on the KXStudio repositories.
Everything working nicely so far.

### DIFF
--- a/linux/JackLinuxFutex.cpp
+++ b/linux/JackLinuxFutex.cpp
@@ -1,0 +1,243 @@
+/*
+Copyright (C) 2004-2008 Grame
+Copyright (C) 2016 Filipe Coelho
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+*/
+
+#include "JackLinuxFutex.h"
+#include "JackTools.h"
+#include "JackConstants.h"
+#include "JackError.h"
+#include <fcntl.h>
+#include <stdio.h>
+#include <sys/mman.h>
+#include <syscall.h>
+#include <linux/futex.h>
+
+namespace Jack
+{
+
+void JackLinuxFutex::BuildName(const char* client_name, const char* server_name, char* res, int size)
+{
+    char ext_client_name[SYNC_MAX_NAME_SIZE + 1];
+    JackTools::RewriteName(client_name, ext_client_name);
+    if (getenv("JACK_PROMISCUOUS_SERVER")) {
+        snprintf(res, size, "jack_sem.%s_%s", server_name, ext_client_name);
+    } else {
+        snprintf(res, size, "jack_sem.%d_%s_%s", JackTools::GetUID(), server_name, ext_client_name);
+    }
+}
+
+bool JackLinuxFutex::Signal()
+{
+    if (!fFutex) {
+        jack_error("JackLinuxFutex::Signal name = %s already deallocated!!", fName);
+        return false;
+    }
+
+    if (fFlush) {
+        return true;
+    }
+
+    if (! __sync_bool_compare_and_swap(&fFutex->futex, 0, 1))
+    {
+        // already unlocked, do not wake futex
+        if (! fFutex->internal) return true;
+    }
+
+    ::syscall(__NR_futex, fFutex, fFutex->internal ? FUTEX_WAKE_PRIVATE : FUTEX_WAKE, 1, NULL, NULL, 0);
+    return true;
+}
+
+bool JackLinuxFutex::SignalAll()
+{
+    return Signal();
+}
+
+bool JackLinuxFutex::Wait()
+{
+    if (!fFutex) {
+        jack_error("JackLinuxFutex::Wait name = %s already deallocated!!", fName);
+        return false;
+    }
+
+    if (fFutex->needsChange)
+    {
+        fFutex->needsChange = false;
+        fFutex->internal = !fFutex->internal;
+    }
+
+    for (;;)
+    {
+        if (__sync_bool_compare_and_swap(&fFutex->futex, 1, 0))
+            return true;
+
+        if (::syscall(__NR_futex, fFutex, fFutex->internal ? FUTEX_WAIT_PRIVATE : FUTEX_WAIT, 0, NULL, NULL, 0) != 0 && errno != EWOULDBLOCK)
+            return false;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


sletz:
int JackInternalClient::Activate()
{
#ifdef __linux__
    fSynchroTable[fClientControl.fRefNum].MakePrivate(true);
#endif
    return JackClient::Activate();
}

int JackInternalClient::Deactivate()
{
    const int ret = JackClient::Deactivate();
#ifdef __linux__
    fSynchroTable[fClientControl.fRefNum].MakePrivate(false);
#endif
    return ret;
}

This  #ifdef __linux__ code is not so clean, could you possibly rework it so that is does not appear in "common" (= not OS specific) code?

falkTX:
that would require adding 'MakePrivate()' to the JackSynchro class, even though it's only used in linux.
is that ok?

sletz:
OK, please document this specific behavior.

Also: does the futex code also works with other Linux systems ?

falkTX:
what other linux systems? you mean other distros?

I have the same patch running on a ARM linux embed system (see http://wiki.moddevices.com/wiki/MOD_Duo).
That device is one of the reasons I made the patch.
Pretty much every jack client in there is an internal one, the lower dsp load is quite noticeable.

Everyone else is welcome to try the patch.
I did testing of my own first and can confirm everything works.
So... testing welcome :)

sletz:
> Le 13 juil. 2016 à 15:29, Filipe Coelho <notifications@github.com> a écrit :
> 
> what other linux systems? you mean other distress?

I meant : older...

falkTX:
I tried on a 3.4 kernel fwiw.

Ace17:
I'm testing your futex branch, but no luck.
Hre's what I get:
(I added some debug printf )

Server terminal:

```
$ ./bin/64/jackd --name silence -d dummy
jackdmp 1.9.11
Copyright 2004-2016 Grame.
Could not find any internals in /home/ace/source/jack2-falktx/bin/64!
Could not find any internals in driver directory!
### JackLinuxFutex::BuildName name = jack_sem.1000_silence_system
### JackLinuxFutex::Allocate name = jack_sem.1000_silence_system val = 0
### JackLinuxFutex::BuildName name = jack_sem.1000_silence_freewheel
### JackLinuxFutex::Allocate name = jack_sem.1000_silence_freewheel val = 0
### JackLinuxFutex::BuildName name = jack_sem.1000_silence_jack_simple_client
### JackLinuxFutex::Allocate name = jack_sem.1000_silence_jack_simple_client val = 0
ClientNotify fails name = system notification = 0 val1 = 0 val2 = 0
NotifyAddClient new_client fails name = jack_simple_client
Cannot notify add client
Cannot create new client
```
Client terminal:
```
$ ls -1 /dev/shm
jack-1000-0
jack-1000-1
jack_sem.1000_silence_freewheel
jack_sem.1000_silence_system
jack-shm-registry
jack_silence_1000_0

$ JACK_DEFAULT_SERVER=silence ./bin/64/example-clients/jack_simple_client
Connect: can't connect named semaphore name = jack_sem.1000_silence_system err = No such file or directory
Cannot read socket fd = 5 err = No such file or directory
CheckRes error
JackSocketClientChannel read fail
Cannot open jack_simple_client client
jack_client_open() failed, status = 0x21
```

Any ideas?

falkTX:
try it without the custom server name first.

I never use it myself so I probably missed something needed for it.

Ace17:
Good catch ; it works if I use the default server name.

Ace17:
By increasing the buffer size, I'm able to deadlock some SDL games at exit (when the audio is closed).
$ jack_bufsize 2048

falkTX:
and without the patch that doesn't happen?

Ace17:
My mistake ; it's unrelated to your patch, as I'm able to reproduce the issue with vanilla libjack binaries.

linux/JackLinuxFutex.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    if ((fSharedMem = shm_open(fName, O_CREAT | O_RDWR, 0777)) < 0) {
+        jack_error("Allocate: can't check in named futex name = %s err = %s", fName, strerror(errno));
+        return false;
+    }
+
+    ftruncate(fSharedMem, sizeof(int));
+
+    if ((fFutex = (int*)mmap(NULL, sizeof(int), PROT_READ|PROT_WRITE, MAP_SHARED|MAP_LOCKED, fSharedMem, 0)) == NULL) {
+        jack_error("Allocate: can't check in named futex name = %s err = %s", fName, strerror(errno));
```
mseeber: duplicated error message

linux/JackLinuxFutex.cpp
```diff
@@ -0,0 +1,204 @@
...
+        return true;
+    }
+
+    if ((fSharedMem = shm_open(fName, O_RDWR, 0)) < 0) {
+        jack_error("Connect: can't connect named futex name = %s err = %s", fName, strerror(errno));
+        return false;
+    }
+
+    if ((fFutex = (int*)mmap(NULL, sizeof(int), PROT_READ|PROT_WRITE, MAP_SHARED|MAP_LOCKED, fSharedMem, 0)) == NULL) {
+        jack_error("Connect: can't connect named futex name = %s err = %s", fName, strerror(errno));
```
mseeber: duplicated error message

falkTX:
I've reworked the patch, will submit an updated version soon.

jan-lugfl:
Any news on the rework ?
I used this patch since July last year and did not really encounter any issue until recently.
Adding an additional ALSA card is broken when the patch is applied. No sound is read or written to/from the card and jackd logs tons of "consumer too slow", "ReadResample error" and "PushAndPull ringbuffer failure... reset". So I assume the locking is broken in that case.

jan-lugfl:
I think I isolated the problem I am having here..

Seem like the issue is related to the private flagged Futex' used by jack_load as internal client in combination with alsa-jack.

It seems like the alsa-jack end is blocking the processing of jack_load threads as soon as a port from alsa-jack port is connected to one of jack_load as its a different process and can not unlock the Futex.

To test it, I tried turned the internal Flag for Allocate() off in JackEngine::ClientInternalOpen() which solved my problem. Im going to test the change in the next weeks so see if it does have other nevative side effects.

falkTX:
that's a known "problem", due to optimizations the internal clients can't connect directly to external ones.
there's some tricks for this, but I want to revise the patch soon which will disable these optimizations by default.

falkTX:
I've updated the PR with a single commit, and disabled the optimizations for internal clients for now.
The code needed for them is still there, to make it easier for the next (optional) PR to enable them.

KXStudio users have been running a version of jack2 with this patch applied for a few months already.
I can't guarantee it works for everyone, but for my personal uses and KXStudio/Debian/Ubuntu it does nicely.

adiknoth:
I'm merging this. We can always undo it if it really breaks stuff.

PR #228, opened 2016-08-17
Add autoclose option to jack_load

falkTX:
This PR adds a 'autoclose' option to jack_load.

This is so that the initial process that called jack_load gets terminated if someone else called jack_unload.
Current code was left as-is as much as possible, so we don't change the previous behaviour.

### DIFF
--- a/example-clients/ipload.c
+++ b/example-clients/ipload.c
@@ -32,11 +32,17 @@ char *intclient_name;
 char *load_name;
 char *load_init = "";
 char *server_name = NULL;
+int autoclose_opt = 0;
 int wait_opt = 0;
+volatile int idling = 1;
 
 static void
 signal_handler (int sig)
 {
+	/* do nothing if internal client closed itself */
+	if (idling == 0)
+		return;
+
 	jack_status_t status;
 
 	fprintf (stderr, "signal received, unloading...");
@@ -45,8 +51,24 @@ signal_handler (int sig)
 		fprintf (stderr, "(failed), status = 0x%2.0x\n", status);
 	else
 		fprintf (stderr, "(succeeded)\n");
+	if (autoclose_opt)
+		jack_deactivate(client);
 	jack_client_close (client);
 	exit (0);
+}
+
+static void
+registration_callback (const char *name, int reg, void *arg)
+{
+	if (reg || strcmp(intclient_name, name))
+		return;
+
+	/* this will stop the wait loop and thus close this application. */
+	idling = 0;
+	return;
+
+	/* unused */
+	(void)arg;
 }
 
 static void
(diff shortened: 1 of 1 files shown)


example-clients/ipload.c
```diff
@@ -199,6 +234,9 @@ main (int argc, char *argv[])
 		}
 	}
 
+	if (autoclose_opt)
+		jack_deactivate(client);
+
```
mseeber: Indentation is kind of hard to understand here, maybe add braces and fix indent of the next line (jack_client_close) (Yes this was broken before, but the additional condition makes it even harder to read in my eyes)

mseeber:
Interesting, just some small things for discussion.

- There is a corresponding manpage in `/man/jack_load.0` which would need an update for the additional option.
- wait and autoclose:
  - maybe in the help string change `intclient` to `internal client`, matching the style of the manpage.
  - mention specific behavior of autoclose option (option is ignored when wait is not set)

example-clients/ipload.c
```diff
@@ -45,17 +51,33 @@ signal_handler (int sig)
...
+registration_callback (const char *name, int reg, void *arg)
+{
+	if (reg || strcmp(intclient_name, name))
+		return;
+
+	/* this will stop the wait loop and thus close this application. */
+	idling = 0;
+
+	/* unused */
+	return; (void)arg;
```
mseeber: I am unsure here. Casting to void already is already the signal to the compiler, that  arg is unused so maybe no comment needed? Also if I as a reader did not know that `/* unused*/` refers to `arg` and not the return, I would still wonder about the unreachable code.

Maybe I am getting this totally wrong, is this some trick to please a specific compiler or syntax highlighter?

mseeber:
After thinking for a while, I get the feeling, that maybe this option should be on by default when specifying wait. When thinking even further, is there actually a use case for using the wait option without it? Is having it running in wait even a valid state after another instance of jack_unload has removed the internal client? Maybe it should not be possible to use the wait option without it?

falkTX:
I don't want to change current behaviour or somehow break user scripts that relied on it.
So I made the new behaviour optional.

Making 'autoclose' imply 'wait' is a good idea.

example-clients/ipload.c
falkTX: I was trying to follow the general code style, but it's very inconsistent...
I find that easy to read as-is, but that's me :)

I can add brackets, no problem.

mseeber: That is ok, the main point is the line after the empty line, there is this function call (line 240) that has a totally different indentation. (At least when i look at the source on github)
Maybe you could correct the indent of that line, not adding braces should be OK if that indent is corrected in my eyes.

mseeber:
Yes breaking user scripts is a concern. I was just wondering if they were broken anyway with the old behavior, which does not take into account, that another client can always remove the internal client. Myabe they are not. I won't judge. :wink:

falkTX:
PR updated:
- indentation at the end of the file corrected
- made return/unused comment more clear

mseeber:
The man page part is missing. Other than that, looking good to me.

PR #242, opened 2016-12-19
Toggle between run and stop state

dehnhardt:
There was a thread in the linuxaudio users list about how to toggle the jack transport state between play and stop using the keyboard.
As they were already using jack_transport to set the state but needed to write a script to get the current transport state with jack_showtime I thought there might be a better solution.
Holger

### DIFF
--- a/example-clients/transport.c
+++ b/example-clients/transport.c
@@ -211,6 +211,29 @@ static void com_timeout(char *arg)
 	jack_set_sync_timeout(client, (jack_time_t) (timeout*1000000));
 }
 
+/* Toggle between play and stop state */
+static void com_toggle(char *arg)
+{       
+	jack_position_t current;
+	jack_transport_state_t transport_state;
+
+	transport_state = jack_transport_query (client, &current);
+
+	switch (transport_state) {
+	case JackTransportStopped:
+        	com_play( arg );
+		break;
+	case JackTransportRolling:
+        	com_stop( arg );
+		break;
+	case JackTransportStarting:
+		printf ("state: Starting - no transport toggling");
+		break;
+	default:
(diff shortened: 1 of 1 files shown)


dehnhardt:
That was quick! Thanks Stéphane

example-clients/transport.c
```diff
@@ -211,6 +211,29 @@ static void com_timeout(char *arg)
...
+        	com_play( arg );
+		break;
+	case JackTransportRolling:
+        	com_stop( arg );
+		break;
+	case JackTransportStarting:
+		fprintf(stderr, "state: Starting - no transport toggling");
+		break;
+	default:
+		fprintf(stderr, "state: Starting - no transport toggling");
```
mseeber: Are you sure this is the right message for the default case? Is the default case even reachable and if yes when?

dehnhardt: I'm sure it's not the right message :-(
I dont't know if the default case is reachable, but in general it does any harm (if the message is right)

mseeber: Maybe change the error message to indicate that the default case is an unexpected state. Just to have a different message than the startup message.

dehnhardt: > Maybe change the error message to indicate that the default case is an unexpected state. Just to have a different message than the startup message.

That's what I was doing this moment;-) My question was, weather a default, which might never be reached, can do any harm - I learned to never use a switch statement without having a default case, but that was years ago...
Anyway, my pull request with the changed error message wil follow in a fw minutes.
Thanks for your help

Holger

BTW: Will this kind of changes be synchronized with the jack1 tree or should I create a pull request for jack1 too?

PR #77, opened 2014-11-02
Enhance promiscuous mode

sdrik:

PR #257, opened 2017-03-18
Secure promiscuous mode

sdrik:
This is a rework of PR #77 

Promiscuous mode requires that all clients are launched with a way too permissive 0000 umask.

With this PR, permissions are explicitly set on various shared resources so it is unnecessary to temper with the umask.
Moreover, it is possible to pass a group name or a gid in the JACK_PROMISCUOUS_SERVER environment variable so that the permissions on those resources are restricted to the members of a specific group.

### DIFF
--- a/common/promiscuous.c
+++ b/common/promiscuous.c
@@ -1,0 +1,95 @@
+/*
+ Copyright (C) 2014-2017 Cédric Schieli
+
+ This program is free software; you can redistribute it and/or
+ modify it under the terms of the GNU General Public License
+ as published by the Free Software Foundation; either version 2
+ of the License, or (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program; if not, write to the Free Software
+ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+*/
+
+#ifndef WIN32
+#include <stdlib.h>
+#include <unistd.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <grp.h>
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#endif
+#include "JackError.h"
+#endif
+
+
+int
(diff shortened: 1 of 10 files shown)


twischer-adit:
I have tested this pull request with the following settings
```
$ adduser jack
$ adduser jack-user
$ vi /etc/group
audio:x:29:jack
jack:x:1002:jack-user                                                                                      
jack-user:x:1003:
```
Therefore only the user jack has access to audio (/dev/snd/*).
And all members of group jack (like user jack-user) have access to the jack daemon (/dev/shm/jack*)

```
$ ulimit -r99 -l100000
$ su - jack
jack$ export JACK_PROMISCUOUS_SERVER="jack"
jack$ jackd -d alsa
```

```
$ ulimit -r99 -l100000
$ su - jack-user
jack-user$ export JACK_PROMISCUOUS_SERVER="jack"
jack-user$ aplay -Djack -c2 /dev/urandom
```

It is working without any access right issues.
I tested it on Linux running on an ARMv8.

I would be happy if this PR could be merged in time.
Can I do anything else to speed up the merge of this PR?

falkTX:
I was planning to test this myself
Looking through the changes I don't see anything bad.
Code is clean and can be merged as-is, so I'll do just that :)
Thank you for the pull request.

sdrik:
Thanks !

7890:
@sdrik I'm thinking of adding this to the manpage in the environment section..

The variable name is IMHO misleading, it hints at that the expected value is an instance of "server".
IIUC the value is a group or GID (?). If that's the case, JACK_PROMISCUOUS_GID might be an alternative.

"The environment variable $JACK_PROMISCUOUS_..  takes a group name or GID value. If set, jackd will limit permissions on shared resources to the given group."

^^^ is this a correct abstract?

sdrik:
@7890 The value of the variable is not always a group name/id.
When set to any value, it means the promiscuous mode should be use.
On platforms that support it, if set to a group name or id, it will also limit the permissions on the shared resources to the given group.

So I don't think JACK_PROMISCUOUS_GID would be a good name unless we split the logic between activation (which would keep the original JACK_PROMISCUOUS_SERVER knob) and permission limit. But on platforms that support it, it doesn't really make sense to activate the mode without limiting the permissions...

7890:
@sdrik i understand.. this is not a new variable. It was undocumented and was now extended semantically.
I think a short description of the usage would be well for the manpage, best formulated from someone knowing the details.

sdrik:
@7890 You're right.
I've added a note about the promiscuous mode in my PR #302

7890:
nice

PR #261, opened 2017-04-21
typecast input parameter to int for abs()

kraj:
Fixes
../tests/test.cpp:482:73: error: call of overloaded 'abs(jack_nframes_t)' is ambiguous

because the signature is int abs(int) and its passing
unsigned int to it.

### DIFF
--- a/tests/test.cpp
+++ b/tests/test.cpp
@@ -479,7 +479,7 @@ int process4(jack_nframes_t nframes, void *arg)
 	jack_nframes_t delta_time = cur_time - last_time;
 
 	Log("calling process4 callback : jack_frame_time = %ld delta_time = %ld\n", cur_time, delta_time);
-	if (delta_time > 0  && (jack_nframes_t)abs(delta_time - cur_buffer_size) > tolerance) {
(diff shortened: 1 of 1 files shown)


tests/test.cpp
```diff
@@ -479,7 +479,7 @@ int process4(jack_nframes_t nframes, void *arg)
 	jack_nframes_t delta_time = cur_time - last_time;
 
 	Log("calling process4 callback : jack_frame_time = %ld delta_time = %ld\n", cur_time, delta_time);
-	if (delta_time > 0  && (jack_nframes_t)abs(delta_time - cur_buffer_size) > tolerance) {
+	if (delta_time > 0  && (jack_nframes_t)abs(int(delta_time - cur_buffer_size)) > tolerance) {
```
kmatheussen: Not sure about that fix. To me it looks like there's a bug in this code. What if 'cur_buffer_size' is smaller than 'delta_time'?

kraj: @kmatheussen it seems you have a point here. The fix does not address that part, it just makes the types consistent.

falkTX:
This has been solved in #277.

PR #265, opened 2017-04-27
alsa_in.c alsa_out.c

fjapinteric:
Added -S option and supporting code to permit attaching to a named jack2 server.

### DIFF
--- a/example-clients/alsa_in.c
+++ b/example-clients/alsa_in.c
@@ -599,6 +599,7 @@ void printUsage() {
 fprintf(stderr, "usage: alsa_out [options]\n"
 		"\n"
 		"  -j <jack name> - client name\n"
+		"  -S <server name> - server to connect\n"
 		"  -d <alsa_device> \n"
 		"  -c <channels> \n"
 		"  -p <period_size> \n"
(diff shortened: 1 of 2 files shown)


falkTX:
there's a minor thing about spaces vs tabs, but the rest seems fine to me.
I'll let others comment as well.

kmatheussen:
Looks fine to me. Also looked at the source code of jack to make sure it ignores an extra NULL argument.

example-clients/alsa_in.c
```diff
@@ -683,6 +686,10 @@ int main (int argc, char *argv[]) {
 	    case 's':
 		smooth_size = atoi(optarg);
 		break;
+		case 'S':
```
falkTX: indentation on these lines do not match the rest of the file.

example-clients/alsa_out.c
```diff
@@ -686,6 +688,10 @@ int main (int argc, char *argv[]) {
 	    case 's':
 		smooth_size = atoi(optarg);
 		break;
+		case 'S':
```
falkTX: same here, indentation is off.

7890:
I will merge this PR and fix the indentation manually.

PR #270, opened 2017-06-14
Fix some issues on Windows.

y-fujii:
- Make locate_dll_driver_dir() independent from the dll file name.
- Add setenv() and unsetenv(), which are used in JackProxyDriver.cpp (Windows does not have them).
- Miscellaneous fixes.

I've tested this patch on MSYS2, with prebuilt PortAudio MSYS2 package, which supports WASAPI.

The PortAudio MSYS2 package seems to use these patches:
<https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-portaudio>

### DIFF
--- a/common/wscript
+++ b/common/wscript
@@ -142,6 +142,7 @@ def build(bld):
     clientlib.defines = 'HAVE_CONFIG_H'
     clientlib.use = uselib
     if bld.env['IS_WINDOWS']:
+         clientlib.env['cxxshlib_PATTERN'] = 'lib%s.dll'
          clientlib.install_path = '${BINDIR}'
     else:
          clientlib.install_path = '${LIBDIR}'
@@ -208,6 +209,7 @@ def build(bld):
     serverlib.target       = 'jackserver'
     serverlib.use = uselib
     if bld.env['IS_WINDOWS']:
+         serverlib.env['cxxshlib_PATTERN'] = 'lib%s.dll'
          serverlib.install_path = '${BINDIR}'
     else:
          serverlib.install_path = '${LIBDIR}'
@@ -303,6 +305,7 @@ def build(bld):
         netlib.target       = 'jacknet'
         netlib.use = ['SAMPLERATE', 'CELT', 'OPUS', 'PTHREAD']
         if bld.env['IS_WINDOWS']:
+             netlib.env['cxxshlib_PATTERN'] = 'lib%s.dll'
              netlib.install_path = '${BINDIR}'
              netlib.use += ['WS2_32', 'WINMM']
         elif bld.env['IS_MACOSX']:
--- a/windows/portaudio/JackPortAudioDriver.cpp
+++ b/windows/portaudio/JackPortAudioDriver.cpp
@@ -196,8 +196,12 @@ int JackPortAudioDriver::Open(jack_nframes_t buffer_size,
     }
 
     // Generic JackAudioDriver Open
+    char capture_driver_name[JACK_CLIENT_NAME_SIZE];
+    char playback_driver_name[JACK_CLIENT_NAME_SIZE];
+    snprintf(capture_driver_name, sizeof(capture_driver_name), "%s", capture_driver_uid);
+    snprintf(playback_driver_name, sizeof(playback_driver_name), "%s", playback_driver_uid);
     if (JackAudioDriver::Open(buffer_size, samplerate, capturing, playing, inchannels, outchannels, monitor,
-        capture_driver_uid, playback_driver_uid, capture_latency, playback_latency) != 0) {
+        capture_driver_name, playback_driver_name, capture_latency, playback_latency) != 0) {
         return -1;
     }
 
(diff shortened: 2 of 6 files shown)


common/wscript
```diff
@@ -142,6 +142,7 @@ def build(bld):
     clientlib.defines = 'HAVE_CONFIG_H'
     clientlib.use = uselib
     if bld.env['IS_WINDOWS']:
+         clientlib.env['cxxshlib_PATTERN'] = 'lib%s.dll'
```
lilrc: What happens if you put `bld.env['cxxshlib_PATTERN'] = 'lib%s.dll'` in the `if bld.env['IS_WINDOWS']` block just above and drop this and similar assignments in the `IS_WINDOWS` block below?

kmatheussen:
Sorry, I didn't see this pull request before creating #271.

We have both implemented setenv/unsetenv, but in different ways. Seems like your way is better. I didn't know about _putenv_s. Instead of using cassert, you could add this line instead:
```c++
if (overwrite==0 && getenv(name)!=NULL)  return 0;
```

The other common thing is the fix for winmme_driver_src in wscript.

common/wscript
y-fujii: It changes some filenames; jack/profiler.dll to jack/libprofiler.dll and jack/netmanager.dll to jack/libnetmanager.dll.
Which is better?

Apart from it, I noticed that my patch has two problems:
1. Calling FreeLibrary() in locate_dll_driver() should be removed.
2. Windows audio device name sometimes longer than JACK_CLIENT_NAME_SIZE. It causes a device name cannot be specified.

In my understanding, changing JACK_CLIENT_NAME_SIZE breaks ABI compatibility. So I changed to use long name in the small part of PortAudio specific code and use truncated name for most common Jack code.

I'm sorry for insufficient tests. I'll fix the pull request.

lilrc: Ah, okay. Then do as you did above, so that the dll-names remain the same.

y-fujii:
The pull request is updated.

- Remove invalid `FreeLibrary()` calls from `locate_dll_driver_dir()`.
- Apply @kmatheussen's suggestion about `setenv()`.
- Change `JackPortAudioDriver::Open()` to permit a long device name and use the name truncated to `JACK_CLIENT_NAME_SIZE` for the most other parts (the 3rd commit).

I just removed `strcpy(f*DriverName, c*_driver_uid);` from `JackPortAudioDriver::Open()` rather than modify them because `JackDriver::Open()` does the same, they seem redundant.

Although it works well on my tests,  I might miss the influence of the 3rd commit.

@kmatheussen,
Thank you for your suggestion.  I applied it as described above.

Your pull request is still worth merging even if mine is merged, of course.  Could you drop the commits 92bcac4 and 34a5e6a from your pull request to make both our patches applicable?

windows/portaudio/JackPortAudioDriver.cpp
```diff
@@ -196,8 +196,12 @@ int JackPortAudioDriver::Open(jack_nframes_t buffer_size,
     }
 
     // Generic JackAudioDriver Open
+    char capture_driver_name[JACK_CLIENT_NAME_SIZE];
+    char playback_driver_name[JACK_CLIENT_NAME_SIZE];
+    snprintf(capture_driver_name, sizeof(capture_driver_name), "%s", capture_driver_uid);
+    snprintf(playback_driver_name, sizeof(playback_driver_name), "%s", playback_driver_uid);
```
falkTX: from what I know of snprintf usage, if the string is bigger than JACK_CLIENT_NAME_SIZE here, the buffer will not include the last null byte to end the string.
I suggest using `sizeof(...)-1`, then `capture_drive_name[JACK_CLIENT_NAME_SIZE-1] = 0;` afterwards

y-fujii: Thank you for your comment.

After the survey, I understand that this problem is more complicated than I thought... Let me summarize it (related to #182):

- In the specification, `snprintf()` always generates null-terminated results.
- `snprintf()` behaves correctly in most platforms, including MSVC >= 2015 and MinGW64.  MinGW64 uses msvcrt.dll but has own implementation.
- MSVC < 2015 does not have `snprintf()` but `_snprintf()`, which does NOT always generate null-terminated results.

and Jack source code does:

- `#define snprintf _snprintf` when the compiler is MSVC.
- The platform independent code seems to assume that `snprintf()` generates null-terminated results.
- The Windows dependent code seems to be mixed.  Some assumes null-terminated, and the other assumes non-null-termimanted.

So my proposal is: 

- Just remove `#define snprintf _snprintf` or make a thin wrapper for MSVC < 2015 to conform the specification.

- Eventually fixes the code that assumes non-null-terminated (Assuming non-null-terminated does not cause serious bugs in most cases, but result length may be shorter than expected).

lilrc:
`man snprintf` says:
```
The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.
```
so any standard conforming (C99, POSIX.1-200{1,8}) implementation of `snprintf` must null terminate the string.

I suggest adding a test in `wscript`, and write a wrapper if `snprintf` misbehaves.

y-fujii:
Implement `_snprintf()` wrapper.

kmatheussen:
> Your pull request is still worth merging even if mine is merged, of course. Could you drop the commits 92bcac4 and 34a5e6a from your pull request to make both our patches applicable?

@y-fujii 
Sure, but maybe you could separate out these two simple issues from this pull request and make a new pull request? I don't want things to break from my tree or pull request, which would happen if I removed those two commits. (Another thing, if you had made separate commits for these things for this pull request, I could have simply dropped my commits, and then cherry-picked your commits into my tree/pull request.)

kmatheussen:
> Another thing, if you had made separate commits for these things for this pull request, I could have simply dropped my commits, and then cherry-picked your commits into my tree/pull request.

In fact, you can do that with my commits. Just cherry-pick the two commits into this tree, fix conflicts (i.e. remove most of my code), and then both our pull requests should be compatible.

falkTX:
I don't want to get this be open forever, might break #271 a little, so that PR will have to be updated.

PR #272, opened 2017-06-29
typecast input parameter to int for abs() to fix build with gcc7

schnitzeltony:
Fixes
../tests/test.cpp:482:73: error: call of overloaded 'abs(jack_nframes_t)' is ambiguous

because the signature is int abs(int) and its passing
unsigned int to it.

Signed-off-by: Andreas Müller <schnitzeltony@googlemail.com>


schnitzeltony:
OOps - Khem did already send this - sorry

PR #273, opened 2017-07-05
Fix unused{,-but-set}-variable compiler warnings.

karllinden:
Since I am mostly hacking in the build system I thought it would be a good idea to let you review these changes in a pull request before pushing to master.

This fixes some (I am not compiling all features so I assume not all) unused-variable and unused-but-set-variable compiler warnings. The fixes are naive, i.e. the variables have been dropped without further ado.


karllinden:
This has been open for more than a year now. Merging.

PR #276, opened 2017-07-13
Fast loading of internal clients and connections from a configuration file

twischer-adit:
This patch introduces a very fast way to load internal clients and connections on start up of the JACK daemon.
It is required on our systems to perform a very fast system start up.

An example configuration file looks like the following:
l inprocess1 inprocess
load amp1 jalv http://lv2plug.in/plugins/eg-amp
c amp:out system:playback_1

An example command line could look like the following:
$ jackd -C /etc/jack_paths.conf -d alsa

### DIFF
--- a/common/JackLoader.cpp
+++ b/common/JackLoader.cpp
@@ -1,0 +1,148 @@
+#include <fstream>
+#include "JackLoader.h"
+#include "JackLockedEngine.h"
+
+
+namespace Jack
+{
+
+JackLoader::JackLoader(JackServer* const server) :
+    fServer(server)
+{
+}
+
+int JackLoader::Load(const std::string file)
+{
+    std::ifstream infile(file);
+
+    if (!infile.is_open()) {
+        jack_error("JACK configuration file %s does not exist or cannot be opened for reading.", file.c_str());
+        return -1;
+    }
+
+
+    std::string line;
+    int linenr = -1;
+    while (std::getline(infile, line))
+    {
+        linenr++;
+
+        std::istringstream iss(line);
+
+        std::string command;
+        if ( !(iss >> command) ) {
+            /* ignoring empty line or line only filled with spaces */
+            continue;
+        }
+
+        /* convert command to lower case to accept any case of the letters in the command */
+        std::transform(command.begin(), command.end(), command.begin(), ::tolower);
+
+        if ( (command.compare("c") == 0) || (command.compare("connect") == 0) ) {
+            ConnectPorts(iss, linenr);
+        } else if ( (command.compare("l") == 0) || (command.compare("load") == 0) ) {
+            LoadClient(iss, linenr);
+        } else if (command.front() == '#') {
+            /* ignoring commented lines.
+             * The # can be followed by non spaces.
+             * Therefore only compare the first letter of the command.
+             */
+        } else {
+            jack_error("JACK configuration file %s line %u contains unkown command '%s'. Ignoring the line!", file.c_str(), linenr, line.c_str());
+        }
+    }
+
+    return 0;
+}
+
+void JackLoader::LoadClient(std::istringstream& iss, const int linenr)
+{
+    std::string client_name;
+    if ( !(iss >> client_name) ) {
+        jack_error("Cannot read client name from configuration file line %u '%s'. Ignoring the line!", linenr, iss.str().c_str());
+        return;
+    }
+
+    std::string lib_name;
+    if ( !(iss >> lib_name) ) {
+        jack_error("Cannot read client library name from configuration file line %u '%s'. Ignoring the line!", linenr, iss.str().c_str());
+        return;
+    }
+
+    /* get the rest of the line */
+    std::string parameters;
+    if ( std::getline(iss, parameters) ) {
+        /* remove the leading spaces */
+        const std::size_t start = parameters.find_first_not_of(" \t");
+        if (start == std::string::npos) {
+            /* Parameters containing only spaces.
+             * Use empty parameter string.
+             */
+            parameters = "";
+        } else {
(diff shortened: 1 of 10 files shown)


common/JackLoader.cpp
```diff
@@ -0,0 +1,129 @@
...
+        jack_error("could not load %s, status = 0x%2.0x",
+                 client_name.c_str(), status);
+        return;
+    }
+    if (status & JackNameNotUnique) {
+        jack_error("Internal client name `%s' not unique",
+                 client_name.c_str());
+        return;
+    }
+
```
kmatheussen: Why not catch all situations where status != 0?

Also, doesn't fServer->InternalClientLoad1 always return false when status != 0?

common/Jackdmp.cpp
```diff
@@ -202,6 +202,7 @@ static void usage(FILE* file, jackctl_server_t *server, bool full = true)
             "               [ --port-max OR -p maximum-number-of-ports]\n"
             "               [ --slave-backend OR -X slave-backend-name ]\n"
             "               [ --internal-client OR -I internal-client-name ]\n"
+            "               [ -C configuration-file ]\n"
```
kmatheussen: All the other options have long and short versions. Maybe the -C option also should have a long version.

common/JackLoader.cpp
```diff
@@ -0,0 +1,129 @@
...
+    int linenr = -1;
+    while (std::getline(infile, line))
+    {
+        linenr++;
+
+        std::istringstream iss(line);
+
+        char command = 0;
+        if ( !(iss >> command) ) {
+            /* ignoring empty line */
```
kmatheussen: Does it also ignore lines with whitespace only?

kmatheussen:
Would it make sense to use "Load" instead of "l", and "Connect" instead of "c"? It's a little bit hard to see those small "l" and "c" characters.

kmatheussen:
Maybe a more specific name than "Configuration" should be used. It could be confused with the .jackd configuration file.

twischer-adit:
@kmatheussen 
> Would it make sense to use "Load" instead of "l", and "Connect" instead of "c"? It's a little bit hard to see those small "l" and "c" characters.

I would like to have the commands as short as possible because the arguments of the internal clients can be very long. But it would be fine for me to support `l`, `c`AND `load`, `connect` as valid commands.


> Maybe a more specific name than "Configuration" should be used. It could be confused with the .jackd configuration file.

I think you mean the `.jackdrc`, `/etc/jackdrc` and `/etc/jackd.conf` which can be used to define the default arguments list of the JACK daemon, right?
Some example names which came into my mind:
- audio-path-configuration
- audio-route-configuration
Have you possibly better name?

common/Jackdmp.cpp
twischer-adit: What do you think about `audio-path-conf`?

common/JackLoader.cpp
twischer-adit: I have verified it. The following call stack is used.
```
JackServer::InternalClientLoad1()
JackServer::InternalClientLoadAux()
JackServer::InternalClientLoad1()
JackInternalClient::Open()
JackInternalClientChannel::ClientCheck()
JackEngine::ClientCheck()
```
All of these functions are returning a negative value when `status` was changed to non-zero.
Therefore I only evaluate the `status` to find the reason for a failure, now.

twischer-adit:
@kmatheussen 
I could not found your comment anymore. Possible you have already found the answer.
> Does it also ignore lines with whitespace only?

I have tested it with spaces and tabs and it is ignoring the whole line without any error messages.

kmatheussen:
First, I forgot to tell write I'm not a maintainer. I just had some time and looked quickly over some pull requests. But I would think/hope @falkTX would have pointed out the same things.

> I would like to have the commands as short as possible because the arguments of the internal clients can be very long. But it would be fine for me to support l, cAND load, connect as valid commands.

Yeah, it was just a suggestion. There might be other pros and cons are here.

> I think you mean the .jackdrc, /etc/jackdrc and /etc/jackd.conf which can be used to define the default arguments list of the JACK daemon, right?

Yes.


>Some example names which came into my mind:
>
>    audio-path-configuration
>    audio-route-configuration
> 
> Have you possibly better name?

Yes, both of these sound better.


> I could not found your comment anymore.

Outdated comments are hidden when you force-push. Search for "Show outdated".

common/Jackdmp.cpp
falkTX: Configuration is really a bad name here.
I'm not a native english speaker, so thinking of names is something I know I'm bad at...

twischer-adit: I changed it to `audio-path-conf`.

falkTX:
I'd like to merge this, but I still don't agree with the names used here.
It's not an audio configuration, but a list of clients and connections..
JackLoader as class name is very generic (I know the current ones in the codebase are not that better..)

@twischer-adit do you mind if I import your changes manually, adjusting just the overall names used?

twischer-adit:
@falkTX 
> do you mind if I import your changes manually, adjusting just the overall names used?

I am fine with it. Feel free to change the names.
Give me a ping if you need any action from me.

falkTX:
ok, I've commited your changes into a new `internal-session-file` branch.
take a look and let me know what you think.
I haven't tried this just yet, will do it soon.

twischer-adit:
I am fine with the changes of https://github.com/jackaudio/jack2/commit/2ad48419a7012b35e5e394530638686c825be4ba.

falkTX:
Alright, merged into master just now.
Thanks for the initial patch, it's quite a handy feature!

PR #277, opened 2017-07-15
Fix gcc7 build and JackDebugClient

kmatheussen:


falkTX:
All the commits seem ok, but you pushed unrelated stuff to this PR to your fork/branch.
I'll rename the PR and merge

tests/test.cpp
```diff
@@ -479,7 +479,7 @@ int process4(jack_nframes_t nframes, void *arg)
 	jack_nframes_t delta_time = cur_time - last_time;
 
 	Log("calling process4 callback : jack_frame_time = %ld delta_time = %ld\n", cur_time, delta_time);
-	if (delta_time > 0  && (jack_nframes_t)abs(delta_time - cur_buffer_size) > tolerance) {
+	if (delta_time > 0  && abs((int64_t)delta_time - (int64_t)cur_buffer_size) > (int64_t)tolerance) {
```
kraj: I think abs() operates on int type. You might want to use llabs() probably if you want to use 64bit type

kmatheussen: In C++, there are both abs(int 32) and abs(int 64) versions.

kraj: ah yes, I forgot it was c++

PR #281, opened 2017-07-20
Be more verbose about $JACK_NO_AUDIO_RESERVATION

7890:
inspired by discussion on [Jack-Devel] Jack won't start on boot


falkTX:
Seems harmless and useful at the same time, thanks!

7890:
thanks

PR #282, opened 2017-07-22
waf: Refactor auto options into a submodule.

### LINKED ISSUE #176: waf: warnings when opus library is found but not opus_custom.h
Note to self bug.

This results in both -DHAVE_OPUS=1 and -DHAVE_OPUS=0, which is bad.

### LINKED ISSUE #177: waf: refactor AutoOptions
Note to self bug.

These are not yet very robust. Also they should be made possible to load() from a waf context.

karllinden:
* The autooptions code is now maintained separately, and imported into
   the project using a git submodule. This fixes #177.
 * The new autooptions are more robust and wscript has been changed
   accordingly.
 * Doxygen is now disabled by default.
 * This also fixes #176.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "autooptions"]
+	path = autooptions
+	url = https://gitlab.com/karllinden/waf-autooptions.git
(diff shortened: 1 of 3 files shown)


lilrc:
I thought I could make a PR since we are in the middle of a release candidate, and because I don't know I you would like the project to start using submodules (although I find it a good idea). I can fast-forward merge this as soon as you have approved it.

falkTX:
Some time later I want to merge the headers and examples/tools of jack1 and jack2, which will mean using submodules.
Let's leave this open and pending until that happens. Thank you.

karllinden:
Rebased onto current master (cc8576a7a).

karllinden:
All changes of this PR except for moving the code to a submodule has been committed to master. The autooptions can be moved into a submodule in the future.

PR #286, opened 2017-08-07
Allow for effective UID of binaries

l1n:
I was trying to use a setuid jack_lsp and it didn't work, presumably because of this - maybe not the most common use case, but I think its helpful to support.


falkTX:
Does not hurt to have, will merge. Thanks

PR #289, opened 2017-08-16
jack_wait: return exit code failure on timeout

twischer-adit:
Signed-off-by: Timo Wischer <twischer@de.adit-jv.com>

### DIFF
--- a/example-clients/wait.c
+++ b/example-clients/wait.c
@@ -122,10 +122,10 @@ main(int argc, char *argv[])
 			}
 		}
 		if (wait_timeout) {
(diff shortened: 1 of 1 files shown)


falkTX:
the idea is great, but please keep the existing indentation in place.

twischer-adit:
@falkTX
I replaced it with taps like the rest of the source file.
Should I really leave it at is it?
This is the only position in this file where spaces were used to indent.

Alternatively I could create a separate commit for it...

falkTX:
A separate commit is better yes. thanks.

twischer-adit:
Done

falkTX:
Looks good now, thank you!

PR #302, opened 2017-10-29
Add a note about the promiscuous mode

sdrik:


falkTX:
Forgot about this, will merge now, thank you.

Issue #197, opened 2016-03-29
WAF : build on solaris/openindiana


benf33:
Hi,
I'm trying to build jack2 on openindiana/hipster but waf doesn't seem to support solaris like system : 
Openindiana/Hipster
> $ uname -a
> SunOS starsystem 5.11 illumos-5a4ef21 i86pc i386 i86pc
> $ python -c "import sys; print sys.platform"
> sunos5

Solaris 10
> $ uname -a
> SunOS darkstar 5.10 Generic_127127-11 sun4u sparc SUNW,A70
> $ python -c "import sys; print sys.platform"
> sunos5

Solaris 8
> $ uname -a
> SunOS deadstar 5.8 Generic_117350-51 sun4u sparc SUNW,Sun-Blade-2500
> $ python -c "import sys; print sys.platform"
> sunos5

sunos5 isn't recognised by waf.
If I add sunos5 to the IS_SUN variable in wscript, I end up with : 
>   File "/home/ben/tmp/jack/jack-1.9.10/.waf-1.6.11-30618c54883417962c38f5d395f83584/waflib/extras/compat15.py", line 151, in apply_uselib_local
>     self.link_task.set_run_after(y.link_task)
> AttributeError: 'task_gen' object has no attribute 'link_task'

If I add sunos5 to the IS_LINUX variable in wscript, I go way further but : 

> Build failed
>  -> task in 'netone' failed (exit status 1): 
>         {task 4260820716L: cxx JackNetOneDriver.cpp -> JackNetOneDriver.cpp.5.o}
[...]

PR #318, opened 2017-12-30
Improve build on Windows with MSVC

clvn:
Greetings!

These are a series of fixes to eliminate most build errors encountered so far on Windows without any major intrusions. 

I have built almost everything in the repo with Visual Studio 2017 using the "msvc" tool in waf on my machine, but there is some remaining work that may be appropriate to hold off a bit:

### TODO
I'm still building tre myself, and using vcpkg to get the other dependencies (pthreads, portaudio, custom libsamplerate port). And using a custom pkg-config.bat to provide linker arguments to waf. Still sorting out how to improve and document this.

I intend to introduce a new SERVER_STATIC flag in *\JackCompilerDeps_os.h for when building libjack.dll, and changing + fixing the semantics of the SERVER_SIDE flag to properly link drivers against libjackserver.dll.

The `jack_info_callback` and `jack_error_callback` data exports defined in jack.h are not compatible with the definitions in JackError.h. They are not exported from the latest Windows dlls, and so I intend to put these inside an `#ifndef WIN32` guard.

I intend to update the wscripts to handle getopt, SERVER_SIDE/SERVER_STATIC fixme's, linking shell32+advapi, M_PI, min, max.

example-clients/showtime and netmaster need usleep(). May just drop a win32 implementation into each of them. inprocess.c needs declspec(dllexport).

Haven't looked into building 64bit binaries at all yet.

Regards
Anders

### DIFF
--- a/common/JackAudioAdapterFactory.cpp
+++ b/common/JackAudioAdapterFactory.cpp
@@ -52,7 +52,7 @@ extern "C"
 
     using namespace Jack;
 
-    SERVER_EXPORT int jack_internal_initialize(jack_client_t* jack_client, const JSList* params)
+    LIB_EXPORT int jack_internal_initialize(jack_client_t* jack_client, const JSList* params)
     {
         jack_log("Loading audioadapter");
 
@@ -78,7 +78,7 @@ extern "C"
         }
     }
 
-    SERVER_EXPORT int jack_initialize(jack_client_t* jack_client, const char* load_init)
+    LIB_EXPORT int jack_initialize(jack_client_t* jack_client, const char* load_init)
     {
         JSList* params = NULL;
         bool parse_params = true;
@@ -97,7 +97,7 @@ extern "C"
         return res;
     }
 
-    SERVER_EXPORT void jack_finish(void* arg)
+    LIB_EXPORT void jack_finish(void* arg)
     {
         Jack::JackAudioAdapter* adapter = static_cast<Jack::JackAudioAdapter*>(arg);
 
--- a/common/JackDriverLoader.cpp
+++ b/common/JackDriverLoader.cpp
@@ -32,6 +32,10 @@ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #include <dirent.h>
 #endif
 
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+#endif
+
 #ifdef WIN32
 
 static char* locate_dll_driver_dir()
--- a/common/JackDummyDriver.cpp
+++ b/common/JackDummyDriver.cpp
@@ -23,7 +23,9 @@ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #include "JackThreadedDriver.h"
 #include "JackCompilerDeps.h"
 #include <iostream>
+#ifndef WIN32
 #include <unistd.h>
+#endif
 #include <math.h>
 
 
--- a/common/JackDummyDriver.cpp
+++ b/common/JackDummyDriver.cpp
@@ -34,7 +34,7 @@ extern "C"
 {
 #endif
 
-    SERVER_EXPORT jack_driver_desc_t * driver_get_descriptor () {
+    LIB_EXPORT jack_driver_desc_t * driver_get_descriptor () {
         jack_driver_desc_t * desc;
         jack_driver_desc_filler_t filler;
         jack_driver_param_value_t value;
@@ -60,7 +60,7 @@ extern "C"
         return desc;
     }
 
-    SERVER_EXPORT Jack::JackDriverClientInterface* driver_initialize(Jack::JackLockedEngine* engine, Jack::JackSynchro* table, const JSList* params) {
+    LIB_EXPORT Jack::JackDriverClientInterface* driver_initialize(Jack::JackLockedEngine* engine, Jack::JackSynchro* table, const JSList* params) {
         jack_nframes_t sample_rate = 48000;
         jack_nframes_t buffer_size = 1024;
         unsigned int capture_ports = 2;
--- a/common/JackLoopbackDriver.cpp
+++ b/common/JackLoopbackDriver.cpp
@@ -90,7 +90,7 @@ extern "C"
 {
 #endif
 
-    SERVER_EXPORT jack_driver_desc_t * driver_get_descriptor()
+    LIB_EXPORT jack_driver_desc_t * driver_get_descriptor()
     {
         jack_driver_desc_t * desc;
         jack_driver_desc_filler_t filler;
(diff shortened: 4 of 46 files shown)


clvn:
Have addressed most of the remaining work in a separate branch, and it builds for Windows on AppVeyor now:
https://ci.appveyor.com/project/clvn/jack2 [Build status](https://ci.appveyor.com/project/clvn/jack2/branch/ci)

And still builds for Linux and OSX on Travis-CI:
https://travis-ci.org/clvn/jack2

(Although the build for Android isn't tested but should fail, due to lack of checking the SERVER_STATIC flag)

Not sure all these changes are agreeable, but at least demonstrates the viability of building with MSVC on Windows.

falkTX:
Thank you very much.
I don't have MSVC, so I cannot test this.
In any case, I'll later on review the changes and comment on anything that might need tweaking.
When/If everything is sorted out, I'll merge the changes if Linux builds still work as before.

karllinden:
@clvn This is very good news. Thank you for your work. I assume you are talking about your `clvn/ci` branch. Is there a reason for not putting them here for review? I would like to review the build system changes.

@falkTX would a working Windows build with `waf` mean we can prune the CodeBlocks project files. I think I have a PR that is blocked by required updates in our other two build systems (CodeBlocks and XCode).

clvn:
@karllinden  Yes, currently it's the `clvn/ci` branch that builds on AppVeyor. The additional work there is somewhat more intrusive than this PR, so I figured it was better to split the changes into smaller PRs. This PR focuses mostly on #ifdeffing Windows-specific stuff without affecting much else, leaving changes to the build system for later.

I'd be happy to include the additional commits from `clvn/ci` in this PR if you want

karllinden:
@clvn You are probably right. We could get this merged first and then merge the other changes (which I too found more intrusive). I have approved, but I have two minor questions:

1. Doesn't variable length arrays work on Windows? They are mandated by C99 so they should work with a reasonable compiler.
2. `alloca` is unportable in general and is not handled correctly. The man page says that `alloca.h` should be included, but it never is. That should probably be fixed. I have seen issues with `alloca` on Sun as well, but if I recall correctly the issue got stalled for some reason. @falkTX Would we like to add a compatibility header with extra checks in `wscript` instead of `#ifdef`ing everywhere. I can help out with that if necessary. Also does the `alloca.h` header exist on Windows?

karllinden:
@clvn FYI, here is the issue that got stalled: https://github.com/jackaudio/jack2/issues/197

clvn:
MSVC is generally not considered a reasonable C compiler, it supports C90, so no variable length arrays =)
https://docs.microsoft.com/en-us/cpp/c-language/ansi-conformance

_alloca is defined in malloc.h on Windows:
https://msdn.microsoft.com/en-us/library/wb1s57t5.aspx

karllinden:
@clvn Great, then we could add a compatibility header (say `alloca.h`) that merely includes `malloc.h` and redefines it as `_alloca` on windows. Then `#include <alloca.h>` would work as expected on all OS:es. I can do this for you if you want to, but your ff8740f commit would become partly obsolete.

clvn:
I agree cleaning up the alloca defines is a good idea, it's maybe could've been put into JackSystemDeps_os.h which seems to handle these things? At least for the server side, some example clients use alloca too, and probably shouldn't tie into the server headers

karllinden:
@clvn I think because it is so general it should be put in its own header and preferably an own directory (I am considering if `compat/alloca/alloca.h`) would be reasonable. Then we could add `-Icompat/alloca/alloca.h` to the compiler flags to get the header on systems where it is absent.

Also having it in `JackSystemDeps_os.h` would add alloca stuff to almost all files (I guess), whereas splitting it would make it used only where it is actually needed.

This should also make our fix reusable for other projects.

clvn:
@karllinden Sounds good to me. Did you think to make that part of this PR, or separately?

karllinden:
@clvn I think a separate PR is preferable. Depending on when things get merged one of use would have to rebase. Do you mind waiting til tomorrow? I won't have time tonight, unfortunately.

karllinden:
@clvn I created a PR for the reusable `alloca` workaround. Since I do not have Windows, would you like to rebase your branch (don't forget to back up this branch!) on top of `jackaudio:compat` to see if it works as desired?

clvn:
@karllinden The compat branch works. Rebased my ci branch on top of compat, removed the Windows alloca #defines, and it still builds. It rebased cleanly. Haven't push anything from these tests.

I guess, if the compat branch is merged first, then I can rebase this PR and and remove the excess alloca #defines (including the unused ones that slipped through after triple checking at 2 AM :( )

clvn:
@falkTX Poke =) Anything I can do to progress this?

Btw, have submitted both [tre](https://github.com/Microsoft/vcpkg/pull/2557) and [JACK2](https://github.com/Microsoft/vcpkg/pull/2456) to vcpkg. The tre port simplifies bootstrapping the build environment on Windows (future PR!)

falkTX:
I'll see all stuff related jack2 around April, have a few things I want to finish first.

karllinden:
@clvn Are you still interested in this? Then please rebase your branch on current master. The `alloca` related stuff landed today...

justnope:
@karllinden I'm not sure what the proper way for this is, but if @clvn is unable to continue with this work, I would like to take over. Do I issue another PR with the changes I made to clvn's branch or what do you propose?

7890:
Windows binaries can be built on travis (Linux flavor) or any other Linux by using MXE build environment. The details of the produced .dll and .exe have yet to be tested at runtime but overall it seems doable.
There was one requirement for platform specific builds which tells that optimally every bit of the build chain should be freely accessible (freedom of speech, free beer, no mandatory subscriptions to developer programs etc.)
MXE gives us possibly more freedom than appveyor in the long run. Opinions?

justnope:
I've opened another PR #423. I'm not the author so I can't make changes to it. I've tried contacting clvn but they haven't responded so far. That pr is also rebased to master and adds 64bit compilation.
 
You're right that WIN32 shouldn't be used in case of "unistd.h", I normally use _MSC_VER to check if it's being compiled with msvc. This is needed because unistd.h doesn't exist and produces a fatal error. 

I could put the check and include of unistd.h in systemdeps.h and instead of including unistd.h in every file I'll just include systemdeps.h?
In any case, the strcasecmp should definitely be put in systemdeps.h

Would you mind continuing the discussion in the new PR?

clvn:
FYI I'm unable to support this PR in the near term. Apologies. I'm happy if somebody else can use it to land MSVC build support in jack

7890:
@clvn thanks for kicking this off, the work shouldn't be lost.

7890:
Follow-up PR https://github.com/jackaudio/jack2/pull/423

PR #319, opened 2018-01-01
License the project clearly

karllinden:
Currently the project has no clear license file. This is not a good state since it forces users and downstream maintainers to go through the project and find out which licenses the project is licensed under. This PR adds a LICENSE file describing which licenses this project is under, and the actual licenses. Note that the GPL and LGPL states that a copy of the license should be distributed with the project, which has not been the case before, so that is addressed too.

The C and C++ sources all have decent copyright headers. However, that has not been the case for the `wscript` files. Thus, another important change is that these files are explicitly put under the GPL-2. The reason this is important is because the project is essentially triple licensed; GPL-2 (binaries), LGPL-2.1 (library) and BSD-2 (upstream waf build system), so which license were the `wscript` files under from the beginning? There are other people who have changed the `wscript` files so they should in fact be attributed in the copyright headers.

What do you think @falkTX should we send a mail to Jack-Devel asking all contributors to add themselves to the the copyright header. Alternatively I could add them in this PR. This is most likely also a good idea in case someone opposes to have their contribution put under the GPL-2.


falkTX:
GPL license is not named correctly.
The PR has it as LICENSE.LGPL-2.0

falkTX:
I don't think asking every single person for license stuff is a good idea.
Likely we won't be able to reach everyone anyway.
The build system for a project does not actually need to be licensed, but I think if we really have to do it, GPL2 is fine.
The wscript is not used on the final binaries so LGPL for the jack libs still works fine with this.

karllinden:
@falkTX Oops, I have fixed the GPL-2. I also removed the .0 from GPL-2.0.

Okay, then because the build system had no license maybe it is fine to license it as GPL-2, even in the absence of other contributors.

I think it is still a good idea to have a clear license on the build system. I at least feel confused with no clear definite headers, since there are effectively three licenses floating around in the project. I mean: which license would apply if someone would copy a `wscript` file verbatim?

7890:
It would be fantastic if licensing is reviewed/added (where/if needed) for the next release.

falkTX:
Sorry for such a long time on this issue. I like it overall, but I think the root LICENSE file is confusing/misleading.
JACK2 itself is not available in 3 licenses, rather separate parts of the code are covered in different ones.
libjack (for clients) is LGPL, while libjackserver and jackd is GPL.

nedko:
jackdbus is GPL as well. I'm dealing with this issue in "licenses" branches of the LADI repos for jack2 and jackdbus.

@falkTX  As source distribution via git or tarballs ship wscript and because wscripts are actually programs, although auxilary for jack2&jackdbus themselves, IMO it makes sense to extend licensing for wscripts.

Authorship is trackable via git. So even if maybe being harder to communicate, it will be useful if other authors that contributed to wscript build system give their point of view on the licensing them. For similar issue with cdbus in ladish (but actually also used in jackdbus), @imaami did a gpg-signed message in git tag: https://github.com/LADI/ladish/releases/tag/dbus_helpers_gpl_afl

nedko:
Merged into LADI/jack2 codebase: https://github.com/LADI/jack2/commit/69b04bee68b591e31c39cb5fd59e41390d26e3ff

falkTX:
> Merged into LADI/jack2 codebase: [LADI/jack2@69b04be](https://github.com/LADI/jack2/commit/69b04bee68b591e31c39cb5fd59e41390d26e3ff)

Please stop referring to the ladi forks, it clutters the notifications with no information useful for other developers here.
Thanks

nedko:
@falkTX  The information is provided so to improve collaboration for mutual benefit.

nedko:
Merged into LADI/jackdbus codebase: https://github.com/LADI/jackdbus/commit/c4d05658980bb3ed9a6f33d6746538bb4fe0658d

PR #320, opened 2018-01-01
Introduce "compatibility modules" and handle alloca

karllinden:
This introduces reusable compatibiliy modules. These can be used to work around operating system specifics such as the fact that `alloca.h` is missing on Windows.

In the future one can also implement missing functions in these modules. For example I think `usleep` is missing on windows.


karllinden:
Rebased on current master (a18319571ac0fc1eaa8678e6ad060556e1d6e756).

@falkTX are we ready to merge this?

PR #329, opened 2018-02-11
Midi port name

Houston4444:
I work with alsa_seq midi-driver for a long time and I must say that the main problem with this is the naming of alsa midi ports. On the other hand, it seems to be less CPU intensive than a2j.

Ports have nice aliases but they all are named "system:midi_playback_#" (for outputs, what is wrong) or "system:midi_capture_#" (for inputs).

Hard to see who is my physical midi input, sooperlooper input or myUsbKeyboard input in programs which don't use port aliases in GUI (as Ardour, and many more). Connections are also memorized by programs only with port names, and they have good chances to not be the same each time you use it.

This code will name ports this way: 
system:$CLIENTNAME/midi_$TYPE_$NUM
where CLIENTNAME is name given by client (sooperlooper, USB Keystation...)
$TYPE is capture/playback/in/out according to port type and physical or not
$NUM is the number (Yes Yes!)

I choose to keep them on system: to not disturb clients in them own ports or reserved names.

To set the num, the code check if a port already exists, I don't know if it is the most efficient.

### DIFF
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -666,8 +666,7 @@ LIB_EXPORT int jack_port_rename(jack_client_t* ext_client, jack_port_t* port, co
         jack_error("jack_port_rename called with a NULL port name");
         return -1;
     } else {
-        return 0;
-//         return client->PortRename(myport, name);
+        return client->PortRename(myport, name);
     }
 }
 
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -666,7 +666,8 @@ LIB_EXPORT int jack_port_rename(jack_client_t* ext_client, jack_port_t* port, co
         jack_error("jack_port_rename called with a NULL port name");
         return -1;
     } else {
-        return client->PortRename(myport, name);
+        return 0;
+//         return client->PortRename(myport, name);
     }
 }
 
--- a/linux/.directory
+++ b/linux/.directory
@@ -1,0 +1,4 @@
+[Dolphin]
+Timestamp=2018,2,11,9,33,21
+Version=4
+ViewMode=1
--- a/linux/.directory
+++ b/linux/.directory
@@ -1,4 +1,0 @@
-[Dolphin]
-Timestamp=2018,2,11,9,33,21
-Version=4
-ViewMode=1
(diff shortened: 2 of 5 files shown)


7890:
> Ports have nice aliases but they all are named "system:midi_playback_#" (for outputs, what is wrong) or "system:midi_capture_#" (for inputs).

It's a naming convention and a question of perspective; 'capture' and 'playback' make sense for terminals IMO.

7890:
I'm afraid to say that this PR comes a bit sloppy. It's not pretty clear which problem it solves. I have to close it here. Thanks and please stay tuned!

Houston4444:
Hello ! It's a shame, this PR is almost a year old and I can't remember the way I've done it or how long did it take me.

> this PR comes a bit sloppy.

Could you be more precise ? Code is missing comments ? Something is dangerous ?

>  It's not pretty clear which problem it solves.

This is really clear to me, maybe I misspoke.
By default, alsa_seq midi-driver will name a midi port system:midi_capture_N (if this is an output port) where N is an iterator. 
It is really not practical for saving/restore jack connections because programs can't know which physical or program port it refers. According to the order in which the ports appear, system:midi_capture_2 could refer to my USB keyboard, the MIDI input of my interface, the Sooperlooper input, or anything else.
It's also not practical to use them into programs which don't display aliases (as Ardour, Carla...), because all ports created by this driver will be named system:midi_capture/playback_N.

The idea of this PR is to gives midi ports created by alsa_seq midi driver explicit names.

> It's a naming convention and a question of perspective; 'capture' and 'playback' make sense for terminals IMO.

Agree. But the minimum is to respect the convention applied by JACK. JACK names outputs of physical input ports "capture" and inputs of physical output ports "playback". This is well done by alsa_seq midi-driver for jack port names, but it's reversed for port aliases (alias 1 & 2). There is absolutely no doubt that this is a mistake.

I am open on the fact this PR could be badly coded but I am clearly sure that there is something to rework on this alsa_seq midi driver.

Cheers.

7890:
@Houston4444 the PR just looked like testing a few things out, maybe it was closed too quickly, sorry for any inconvenience. Maybe we split it up to smaller chunks.

> JACK names outputs of physical input ports "capture" and inputs of physical output ports "playback". This is well done by alsa_seq midi-driver for jack port names, but it's reversed for port aliases (alias 1 & 2). There is absolutely no doubt that this is a mistake.

Would you mind to open a separate issue for just this potential bug? With an extensive example what's wrong at which point in time and what you would expect it to be. I agree that the notion of capture and playback should always be crystal clear, and if needed looking at it from a given context/perspective.

PR #344, opened 2018-04-21
Set ToS on Netjack2 packets

jmendeth:
Set the ToS/DS field on the packets sent over the network (the code was commented out for some reason).

This should make most network hops prioritize Netjack2 packets over other kinds of traffic.

### DIFF
--- a/posix/JackNetUnixSocket.cpp
+++ b/posix/JackNetUnixSocket.cpp
@@ -125,21 +125,13 @@ namespace Jack
         0x00 / 000000 / 0	0	Best Effort	BE	AC_BE
         */
         
-        /*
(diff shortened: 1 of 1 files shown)


jmendeth:
Note: the value of `46` means Expedited Forwarding (or for older equipment, critical priority), which is common to use in realtime voice / sound apps.

falkTX:
There must be a reason why the code was disabled before. any ideas on why?

sletz:
Because it has never shown any kind of improvement I guess.

falkTX:
Then I guess it is safe to merge.
We just be on the watch for future bug reports.

PR #348, opened 2018-05-16
Fix detection of USB disconnect for ALSA devices

agoode:
Without this change, jack gets stuck in a tight poll() loop that it never recovers from. This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1416091.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,7 @@ Lars-Peter Clausen
 Alexandru Tofan
 Kim Jeong Yeon
 Filipe Coelho
+Google LLC
 
 ---------------------------
   Jackdmp changes log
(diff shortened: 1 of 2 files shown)


falkTX:
This is something that has been bothering me and I wanted to fix eventually, but have some things to handle first.
So thanks for the fix!
Question.. what does then happen with jack when you remove the soundcard? Do non-audio events still keep happening? Does jack shut down?
I will test this on the weekend anyway.

Last thing, please dont add company names in the ChangeLog.
If you want to, leave your personal name as contributor, not companies.
Thanks.

unfa:
Wow, so if you work for Google, the any FOSS contributions you make in your free time are either credited to Google, or you need to spend extra time filing documents?

That is...

**_special_**

PR #357, opened 2018-07-27
jackdmp: Signal to systemd when ready

twischer-adit:
With this patch it is possibly to start the JACK daemon with a systemd
service file of type notify. The following provides an example service
file:

[Service]
Type=notify
ExecStart=/usr/bin/jackd -d alsa

[Install]
WantedBy=multi-user.target

### DIFF
--- a/common/Jackdmp.cpp
+++ b/common/Jackdmp.cpp
@@ -43,6 +43,10 @@ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #include "audio_reserve.h"
 #endif
 
+#if HAVE_SYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
+
 /*
 This is a simple port of the old jackdmp.cpp file to use the new jack2 control API. Available options for the server
 are "hard-coded" in the source. A much better approach would be to use the control API to:
@@ -624,6 +628,10 @@ int main(int argc, char** argv)
     notify_sent = true;
     return_value = 0;
 
+#if HAVE_SYSTEMD
+    sd_notify(0, "READY=1");
+#endif
+
     // Waits for signal
 #ifdef __ANDROID__
     //reserve SIGUSR2 signal for switching master driver
@@ -637,6 +645,10 @@ int main(int argc, char** argv)
     }
 #else
     jackctl_wait_signals(sigmask);
+#endif
+
+#if HAVE_SYSTEMD
+    sd_notify(0, "STOPPING=1");
 #endif
 
  stop_server:
--- a/wscript
+++ b/wscript
@@ -439,6 +439,9 @@ def options(opt):
     readline = add_auto_option(opt, 'readline', help='Build with readline')
     readline.add_library('readline')
     readline.set_check_hook(check_for_readline, check_for_readline_error)
+    sd = add_auto_option(opt, 'systemd', help='Use systemd notify')
+    sd.add_header('systemd/sd-daemon.h')
+    sd.add_library('systemd')
 
     # dbus options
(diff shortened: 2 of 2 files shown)


dvzrv:
this is very cool! does it work as a systemd --user service, system service, or both?

common/Jackdmp.cpp
```diff
@@ -43,6 +43,10 @@ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #include "audio_reserve.h"
 #endif
 
+#if HAVE_SYSTEMD == 1
```
falkTX: This style of macro is inconsistent with the rest of the codebase.
Use `#ifdef HAVE_SYSTEMD` instead

twischer-adit:
@dvzrv 
> this is very cool! does it work as a systemd --user service, system service, or both?

I am using this feature to start JACK as a system service as a special user.
I am not aware if it can also be used to start JACK as a {{systemd --user}} service.

dvzrv:
if you'd like to try: put the above snippet in `~/.config/systemd/user/jack.service` do a `systemctl --user daemon-reload` and try starting: `systemctl --user start jack`

common/Jackdmp.cpp
twischer-adit: I had some issues when using `#ifdef HAVE_SYSTEMD`. WAF will define `HAVE_SYSTEMD == 0` in case of systemd is not available. See `build/config.h` for example:
```c
...
#define HAVE_CELT 0
#define HAVE_OPUS 0
#define HAVE_SAMPLERATE 1
#define HAVE_SNDFILE 1
#define HAVE_READLINE 1
#define HAVE_SYSTEMD_SD_DAEMON_H 1
#define HAVE_SYSTEMD 1
...
```

The issue is reproducible with the following source code:
```c
#define VAR 0

#ifdef VAR
#error "DEFINED"
#else
#error "UNDEF"
#endif
```
It fails with
```bash
$ gcc main.c
main.c:4:2: error: #error "DEFINED"
 #error "DEFINED"
  ^
```

Therefore I think the other `#ifdef` could also end up in issues.

falkTX: I see, you are correct then.
So just use `#if HAVE_SYSTEMD`. no need for the == 1 comparison.

twischer-adit:
Diff
```patch
diff --git a/common/Jackdmp.cpp b/common/Jackdmp.cpp
index b1d7b39..e7198f4 100644
--- a/common/Jackdmp.cpp
+++ b/common/Jackdmp.cpp
@@ -43,7 +43,7 @@ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #include "audio_reserve.h"
 #endif
 
-#if HAVE_SYSTEMD == 1
+#if HAVE_SYSTEMD
 #include <systemd/sd-daemon.h>
 #endif
 
@@ -628,7 +628,7 @@ int main(int argc, char** argv)
     notify_sent = true;
     return_value = 0;
 
-#if HAVE_SYSTEMD == 1
+#if HAVE_SYSTEMD
     sd_notify(0, "READY=1");
 #endif
 
@@ -647,7 +647,7 @@ int main(int argc, char** argv)
     jackctl_wait_signals(sigmask);
 #endif
 
-#if HAVE_SYSTEMD == 1
+#if HAVE_SYSTEMD
     sd_notify(0, "STOPPING=1");
 #endif
 
-- 
2.7.4
```

falkTX:
ping to @dvzrv to test as user service :)

falkTX:
Tested to work, thank you

PR #358, opened 2018-07-27
alsa driver: Provide option to enable time stamp mode for all devices

twischer-adit:
This is required for example for some adaptive sample rate converters
which are executed on top of JACK. These SRCs are using the time stamp to
calculate the buffer fill level between playback application and sound
card.


falkTX:
Looks good. I trust that it builds and runs as usual if "-t" option is not used.
(that reminds me, I need to add some CI tests here...)

falkTX:
FYI I am removing the option to enable this, and let it be always on.
JACK1 added such code already, and it seems something we should have on by default anyway.

PR #391, opened 2018-11-11
[metadata-remove-property-fix] Fix #390 - Meta-data delete-all crashes the client

### LINKED ISSUE #390: Meta-data delete all crashes the client
```
$ gdb --args jack_property -D
...
Thread 1 "jack_property" received signal SIGSEGV, Segmentation fault.
__strncpy_ssse3 () at ../sysdeps/x86_64/multiarch/strcpy-ssse3.S:43
43      ../sysdeps/x86_64/multiarch/strcpy-ssse3.S: No such file or directory.
(gdb) bt
#0  __strncpy_ssse3 () at ../sysdeps/x86_64/multiarch/strcpy-ssse3.S:43
#1  0x00007ffff7b97301 in ?? () from /usr/lib/x86_64-linux-gnu/libjack.so.0
#2  0x00007ffff7b88dc1 in ?? () from /usr/lib/x86_64-linux-gnu/libjack.so.0
#3  0x00007ffff7bb9686 in ?? () from /usr/lib/x86_64-linux-gnu/libjack.so.0
#4  0x00005555555551ea in ?? ()
#5  0x00007ffff77a9b97 in __libc_start_main (main=0x555555554de0, argc=2, argv=0x7fffffffdcd8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdcc8) at ../csu/libc-start.c:310
#6  0x000055555555568a in ?? ()
```

rncbc:
[metadata-remove-property-fix] Fix #390 - Meta-data delete-all crashes the client.

### DIFF
--- a/common/JackRequest.h
+++ b/common/JackRequest.h
@@ -1647,7 +1647,8 @@ struct JackPropertyChangeNotifyRequest : public JackRequest
     {
         jack_uuid_copy(&fSubject, subject);
         memset(fKey, 0, sizeof(fKey));
-        strncpy(fKey, key, sizeof(fKey)-1);
+        if (key)
+            strncpy(fKey, key, sizeof(fKey)-1);
     }
 
     int Read(detail::JackChannelTransactionInterface* trans)


common/JackRequest.h
```diff
@@ -1647,7 +1647,8 @@ struct JackPropertyChangeNotifyRequest : public JackRequest
     {
         jack_uuid_copy(&fSubject, subject);
         memset(fKey, 0, sizeof(fKey));
-        strncpy(fKey, key, sizeof(fKey)-1);
+        if (key)
+            strncpy(fKey, key, sizeof(fKey)-1);
```
falkTX: now that I look at this... won't it possibly copy more bytes than needed?
if key has length 3 for example, we will copy over some garbage memory.

correct me if I am wrong, but shouldn't the 3rd argument in there instead be something like `min(strlen(key), sizeof(fKey)-1)` ?

rncbc: but that's exactly what strncpy does: the 3rd parameter is the maximum nunber of characters to be copied; strlen(key) gives the number of characters in null-terminated key, which might however be greater than size of the target buffer fKey.

PR #397, opened 2018-11-21
float2int: Use right S24 format and full range S32 format

twischer-adit:
ALSA expects right-aligned (0x00******) samples when using S24_LE and S24_BE but without this PR JACK converts into left-aligned samples (0x******00). This can result in clipping when the amplitude is higher than 0x7FFF.
In addition the same conversion was also used for S32_LE and S32_BE. Therefore the full range of S32_LE was not used. The maximal value of JACK was 0x7FFFFF00 but the maximal value accepted by ALSA is 0x7FFFFFFF. This PR provides separated conversion functions for S32 (inlining is used to reduce duplicates).

I have compiled it for ARM64 and x86_64.

### DIFF
--- a/common/memops.c
+++ b/common/memops.c
@@ -137,17 +137,6 @@
 		(d) = f_round ((s) * SAMPLE_24BIT_SCALING);\
 	}
 
-/* call this when "s" has already been scaled (e.g. when dithering)
- */
-
-#define float_24u32_scaled(s, d)\
-        if ((s) <= SAMPLE_24BIT_MIN_F) {\
-		(d) = SAMPLE_24BIT_MIN << 8;\
-	} else if ((s) >= SAMPLE_24BIT_MAX_F) {	\
-		(d) = SAMPLE_24BIT_MAX << 8;		\
-	} else {\
-		(d) = f_round ((s)) << 8; \
-	}
 
 #define float_24(s, d) \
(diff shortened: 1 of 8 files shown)


twischer-adit:
@falkTX and @dvzrv 
I would expect the clang build error on OSX
```
../posix/JackNetUnixSocket.cpp:164:16: error: no viable conversion from returned value of type '__bind<int &, sockaddr *, unsigned long>' to function return type 'int'
        return bind(fSockfd, reinterpret_cast<socket_address_t*>(&fRecvAddr), sizeof(socket_address_t));
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
is not introduced by this PR. It exists longer but build check was not enabled (https://github.com/jackaudio/jack2/commit/407e8c0513fbf3752b42232e4e8a0dd4fa5f9369) in the past, right?

twischer-adit:
Please review @DeviLaxmii @michael-methner @aditpape

twischer-adit:
@falkTX 
> Can you rebase on master so we can have CI tests?

Done

falkTX:
Thanks a lot!

PR #398, opened 2018-11-21
arm64: Enable simd implementations

twischer-adit:
Without this PR simd optimization will not used for ARMv8.
GCC defines __ARM_NEON__ for asimd on ARMv7 and __ARM_NEON for simd on
ARMv8. Therefore also check for __ARM_NEON.


twischer-adit:
Please review @DeviLaxmii @michael-methner @aditpape

falkTX:
Thanks for spotting, merging now.

Issue #283, opened 2017-07-26
Please provide POSIX fallbacks for all OS-specific implementations


yurivict:
Currently, for example, JackSynchro doesn't have a POSIX implementation. On Linux it is substituted with JackLinuxFutex.

As a result, Jack2 only supports a few systems. Please make sure that it builds on POSIX systems without OS-specific functions. Jack1 works on the FreeBSD, and Jack2 doesn't due to these problems.

In other words, it should build with only "../posix/" includes.

falkTX:
But there is a posix implementation of it already, using semaphores. (sem_init/post/timedwait, etc)
It was changed by me to a futex to work around limitations with semaphores, and also for optimization with internal clients (not enabled by default).

See https://github.com/jackaudio/jack2/blob/master/posix/JackPosixSemaphore.cpp
Probably the wscript just needs an extra rule for freebsd

yurivict:
It doesn't build when I remove "../linux" from the config file. Could you please make sure it builds without "../linux" then?

The error is:
```
/usr/ports/audio/jack2/work/jack2-1.9.11-RC1/common/JackPlatformPlug.h:23:10: fatal error: 'JackPlatformPlug_os.h' file not found
#include "JackPlatformPlug_os.h"
         ^~~~~~~~~~~~~~~~~~~~~~~
```

yurivict:
@falkTX Are you able to remove '../linux' and build on linux?

yurivict:
This is a show-stopper on FreeBSD.

It should build without the "linux" directory. When I tried to build it the POSIX-only build failed.

PR #400, opened 2018-11-23
Add FreeBSD support

unrelentingtech:
(re: #283)

- OSS backend works fine
- ALSA backend is crackling (probably resampling related; that happens other apps too), do not use (wasn't used in jack1 either)
- D-Bus works fine
- tested with Carla, Cadence, qjackctl
- using POSIX semaphores because I'm currently too lazy to write a thing for `umtx` (`futex` equivalent)
  - does any platform need `O_RDWR` in semaphores?
  - would any platform complain about beginning the semaphore path with `/`?
- also needs `-fPIC` to build the OSS adapter, I'm not sure where in waf I should enable PIC

### DIFF
--- a/freebsd/JackAtomic_os.h
+++ b/freebsd/JackAtomic_os.h
@@ -1,0 +1,34 @@
+/*
+Copyright (C) 2004-2008 Grame
+Copyright (C) 2018 Greg V
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+*/
+
+#ifndef __JackAtomic_freebsd__
+#define __JackAtomic_freebsd__
+
+#include "JackTypes.h"
+#include <sys/types.h>
+#include <machine/atomic.h>
+
+static inline char CAS(volatile UInt32 value, UInt32 newvalue, volatile void* addr)
+{
+		 return atomic_cmpset_32((uint32_t*)addr, value, newvalue);
+}
+
+#endif
+
--- a/freebsd/JackPlatformPlug_os.h
+++ b/freebsd/JackPlatformPlug_os.h
@@ -1,0 +1,85 @@
+/*
+Copyright (C) 2004-2008 Grame
+Copyright (C) 2018 Greg V
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program; if not, write to the Free Software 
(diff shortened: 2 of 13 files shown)


freebsd/JackPlatformPlug_os.h
```diff
@@ -0,0 +1,85 @@
...
+#ifndef __JackPlatformPlug_freebsd__
+#define __JackPlatformPlug_freebsd__
+
+#define jack_server_dir "/tmp"
+#define jack_client_dir "/tmp"
+#define JACK_DEFAULT_DRIVER "oss"
+
+namespace Jack
+{
+		struct JackRequest;
```
7890: tab

freebsd/JackPlatformPlug_os.h
```diff
@@ -0,0 +1,85 @@
...
+	struct JackResult;
+		
+		class JackPosixMutex;
```
7890: tab

freebsd/JackAtomic_os.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef __JackAtomic_freebsd__
+#define __JackAtomic_freebsd__
+
+#include "JackTypes.h"
+#include <sys/types.h>
+#include <machine/atomic.h>
+
+static inline char CAS(volatile UInt32 value, UInt32 newvalue, volatile void* addr)
+{
+		 return atomic_cmpset_32((uint32_t*)addr, value, newvalue);
```
7890: tab

7890:
This PR looks good to me, I think we should probably merge it away. Are there any opinions from other FreeBSD users? If possible please also attach a full verbose build log.

freebsd/driver.h
```diff
@@ -0,0 +1 @@
+../linux/driver.h
```
falkTX: We need to figure out something better for this

yurivict:
@myfreeweb Could you please squash all commits into one?

myfreeweb:
@yurivict hmm, why?

@7890 @falkTX can someone please answer the questions from the original post:

> does any platform need `O_RDWR` in semaphores?

> would any platform complain about beginning the semaphore path with `/`?

> I'm not sure where in waf I should enable PIC

yurivict:
> @yurivict hmm, why?

Difficult to add 6 lines, need to maintain the proper order of patches, etc.
I tried to add them into the port but there are patching problems. You probably need to rebase.

Also, ```__FreeBSD_kernel__``` isn't needed.

myfreeweb:
If you want to test in ports, you can [just fetch from my branch](https://github.com/myfreeweb/freebsd-ports-dank/commit/b47eb6d8a3d0ed994897bd394a1e8d548b2fdbb1). If you want to commit a port already… isn't it a bit early? :)

7890:
From `man sem_open`:
```
Both read and write permission should be granted to each class of user that will access  the  semaphore.
```
This hints at that O_RDWR is correct.

From man `sem_overview`:
```
Named semaphores
A named semaphore is identified by a name of the form /somename; that is, a  null-terminated string  of  up to NAME_MAX-4 (i.e., 251) characters consisting of an initial slash, followed by one or more characters, none of which are slashes.
```
Like shared memory handles for shm_open, semaphores should start with a slash. I think this rule isn't strictly enforced, any name without slashes in the middle should work. I speculate that using a leading slash wouldn't be problematic on any platform. However there are different semaphore name length limits on different platforms (see OSX max. client name length).

freebsd/driver.h
7890: I don't understand. Is this about naming, licenses and such or about technical shortcomings?

0EVSG:
@falkTX Could you please elaborate on your review above (`freebsd/driver.h`, Feb 6. 2019)?

If it's about using symlinks, there's the option of a relay `#include` with relative path.

Otherwise, it's probably also acceptable to drop the ALSA backend on FreeBSD, which refers to `driver.h`. As for `uptime.h`, either just duplicate it (it's small), or add the linux directory to the include paths when compiling the DBUS controller?

FYI, jack2 v1.9.16 is now in the FreeBSD ports, taken from a branch in my repo. It is based on these commits plus a separate, overhauled OSS driver taylored to FreeBSD. I intend to make an up-to-date pull request for that, thus my interest in how to fix this.

yurivict:
FYI: Jack2 is now available in FreeBSD ports: https://www.freshports.org/audio/jack/

0EVSG:
Closing this, superseded by #800 which includes most of the work in this pull request.
Thanks everybody for your contributions!

PR #409, opened 2019-01-16
Connect all ports with the same alias

twischer-adit:
If there are multiple ports with the same alias the user has to call jack_connect() only once and all ports with the same alias will be connected. With this feature for example a media player does not need to know which JACK clients are interested in its audio data. The media player has only to connect to the right alias and will automatically be connected to all ports with the same alias.

Without this patch only the first register port with the matching name or alias will be connected. The following ports were ignored.

```
$ cat /etc/jackd-session.conf
a alias:media system:playback_1 system:playback_2
$ jackd -C /etc/jackd-session.conf -d alsa
$ jack_lsp -A
system:capture_1
   alsa_pcm:jack_be_in:out1
system:capture_2
   alsa_pcm:jack_be_in:out2
system:playback_1
   alsa_pcm:jack_be_out:in1
   alias:media
system:playback_2
   alsa_pcm:jack_be_out:in2
   alias:media

# only one connect call is required to connect all ports with alias:media
$ jack_connect system:capture_1 alias:media
$ jack_lsp -c
system:capture_1
   system:playback_1
   system:playback_2
system:capture_2
system:playback_1
   system:capture_1
system:playback_2
   system:capture_1
```

This feature is also useful in combination with the ALSA JACK plugin. The ALSA application can playback the same channel to multiple audio ports (by connecting to an alias) or capture from multiple audio ports.

### DIFF
--- a/common/JackEngine.cpp
+++ b/common/JackEngine.cpp
@@ -955,11 +955,30 @@ int JackEngine::CheckPortsConnect(int refnum, jack_port_id_t src, jack_port_id_t
 int JackEngine::PortConnect(int refnum, const char* src, const char* dst)
 {
     jack_log("JackEngine::PortConnect ref = %d src = %s dst = %s", refnum, src, dst);
-    jack_port_id_t port_src, port_dst;
+    jack_port_id_t port_src, first_port_dst;
 
-    return (fGraphManager->GetTwoPorts(src, dst, &port_src, &port_dst) < 0)
-           ? -1
-           : PortConnect(refnum, port_src, port_dst);
+    if (fGraphManager->GetTwoPorts(src, dst, &port_src, &first_port_dst) < 0) {
+        return -1;
+    }
+
+    int ret = 1;
+    /* connect all ports if there are multiple ports with the same alias */
+    for (; port_src != NO_PORT; port_src = fGraphManager->GetNextPort(src, port_src)) {
+        for (jack_port_id_t port_dst = first_port_dst; port_dst != NO_PORT; port_dst = fGraphManager->GetNextPort(dst, port_dst)) {
+            const int res = PortConnect(refnum, port_src, port_dst);
+            if (res < 0) {
+                const char* const src_name = fGraphManager->GetPort(port_src)->GetName();
+                const char* const dst_name = fGraphManager->GetPort(port_dst)->GetName();
+                jack_error("Connecting src_name [%s] (alias [%s]) and dst_name [%s] (alias [%s]) failed!",
+                           src_name, src, dst_name, dst);
+            }
+            if (res < ret) {
+                ret = res;
+            }
+        }
+    }
+
+    return ret;
 }
 
 int JackEngine::PortConnect(int refnum, jack_port_id_t src, jack_port_id_t dst)
--- a/common/JackGraphManager.cpp
+++ b/common/JackGraphManager.cpp
@@ -742,15 +742,25 @@ int JackGraphManager::GetTwoPorts(const char* src_name, const char* dst_name, ja
 }
 
 // Client : port array
-jack_port_id_t JackGraphManager::GetPort(const char* name)
+jack_port_id_t JackGraphManager::GetPortAux(const char* const name, const jack_port_id_t current)
 {
-    for (unsigned int i = 0; i < fPortMax; i++) {
+    for (unsigned int i = current; i < fPortMax; i++) {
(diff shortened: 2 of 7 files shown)


falkTX:
I do not like this idea at all...
We have meta-data now for all clients and ports, so we can instead define a property (which ports are important to connect to, for example) and let applications listen to it if they desire.
Aliases were a mistake, something put in place only because of the lack of a better system.
We really need to move away from them now, no more reason to keep using them.
They will be a deprecated API quite soon.

twischer-adit:
@falkTX 
Thanks a lot for the fast feedback.

> We have meta-data now for all clients and ports, so we can instead define a property

I would expect you are talking about the metadata API https://github.com/jackaudio/jack2/blob/master/common/jack/metadata.h.
I found the definition of `JACK_METADATA_PORT_GROUP`. Therefore I would define an "alias" (aka port group) in the following way:
```
jack_uuid_t uuid = jack_port_uuid(port);
jack_set_property(client, uuid, JACK_METADATA_PORT_GROUP, "[NAME_OF_THE_GROUP]", "text/plain");
```

So an application could automatically connect to all ports of the same group. But for me it looks like I have to iterate over all subjects (aka ports) to find the port groups. Or did I miss any search API? I only saw `jack_get_properties(jack_uuid_t subject, jack_description_t* desc)` but I have to call it for each port.


Has this feature to be implemented for each JACK client or could I add this feature to the `jack_connect()` call? For example `jack_connect("system:capture_1", "port-group:media")` will automatically connect `system:capture_1` to all ports of the group `media`.

7890:
Connecting to an alias port which abstracts a group of ports involves an unknown number of ports. I am undecided about whether a client or the server can handle that case better. Thinking naively, user connects to port "default" and without user doing there are many "default" aliases. Would jack_connect() hit critical timing limits in the loop?

Related programs jack_capture and jack_plumbing come to mind which can handle connections dynamically.

The meta-data API allows a more fine-grained, fully user-controlled model of the jack objects. A smart client can connect according to user-defined meta-data dependent rules. The full potential of the meta-data API hasn't yet been utilised since it's a relatively new addition. It could almost certainly be useful for smart connection scenarios.

twischer-adit:
@7890 
> Would jack_connect() hit critical timing limits in the loop?

As far as I got all `jack_connect()` calls are synchronized with the mutex used by `JackLockedEngine`. But this mutex will only be locked for JACK requests. Therefore it will not be locked for usual audio processing. So the audio processing will not be blocked by a long time `jack_connect()` call.

Anyway you are right accessing the DB of properties should not block the mutex. A good solution could be an Implementation of a calls which inherits from `JackLockedEngine`. This class could implement any automatic rules and features you like. For this PR I would overwrite `PortConnect()` and iterate over all ports if a port is given starting with `port-group:`.
So I could use the same implementation for the internal session manager (`-C` option) and all JACK clients) Non of them has to be changed. The port iteration would not be executed without JACK IPC mechanism in between (less CPU overhead) but the mutex will only be locked when calling `PortConnect()` but not when calling `jack_get_properties()`.

What do you think?

7890:
I think it has some merit, but my insights are limited. If alias ports are going to be deprecated in favour of the meta-data API (see @falkTX comment) using alias ports seems like a one-way route. Changing internal semantics is always a hot iron, if the use case can be implemented less intrusive (eg. a client-only solution), it would certainly be preferred. Please hold on to get more feedback from @falkTX

twischer-adit:
The solution described by me in https://github.com/jackaudio/jack2/pull/409#issuecomment-455082391 and https://github.com/jackaudio/jack2/pull/409#issuecomment-456006619 does also not full fill my requirements.

My idea was to define the groups on JACK startup only once. But a JACK port cannot be added to a group (the property cannot be defined) as long as the port is not registered. In addition the property has to be set again if the port was unregistered and registered again (because the port number could change). Therefore I would need an additional JACK client which is listing on port registrations and set the properties accordingly to a configuration file. But this would increase the processing overhead and is not really acceptable for my embedded system.

The main reason for this PR was the feature to connect the same ALSA channel (left, right, ...) to multiple JACK ports when using the ALSA JACK plugin. Therefore I adapted the ALSA JACK plugin to keep the processing overhead for the port properties handling low. (See http://mailman.alsa-project.org/pipermail/alsa-devel/2019-January/144632.html)

Thanks for your comments. This helps to find the right solution for my issue.

PR #412, opened 2019-01-17
Protect against real time skips

milloni:
POSIX _timedwait API is affected by changes to system time (that is, a sysadmin/service has changed the datetime on the system). This pull request attempts to mitigate that by additionally measuring elapsed time with CLOCK_MONOTONIC, and restarting the _timedwait call if a time skip had occurred.

I've run into this problem in a client's in-vehicle infotainment system. Note there's also an effort to create a GNU extension API that will function in a similar way as _timedwait, but will let the user specify the clock [1]. By choosing CLOCK_MONOTONIC, the user can avoid this issue.

[1] http://sourceware-org.1504.n7.nabble.com/PATCH-Semaphores-add-sem-timedwait-monotonic-GNU-extension-BZ-14717-td550206.html#a550467


posix/JackPosixSemaphore.cpp
```diff
@@ -115,29 +116,66 @@ bool JackPosixSemaphore::Wait()
...
+        }
+
+        /* Compare with monotonic timeout, in case a step happened */
+        int old_errno = errno;
+        res = clock_gettime(CLOCK_MONOTONIC, &now_mono);
+        assert(res == 0);
+        errno = old_errno;
+        if (JackPosixTools::TimespecCmp(&now_mono, &end_mono) >= 0) {
+            goto err;
+>>>>>>> c2dcdf2... posix: Wrap sem_timedwait() to check for time jumps
```
falkTX: I think you forgot something here, this line should not be here

falkTX:
PR looks good overall.
I thought there was a non-portable call to setup a semaphore clock..

also note that posix-semaphore c++ code is not used under linux, that is why CI passes even though the code itself should not build.

falkTX:
the asserts against clocktime return value seem dangerous to me.
on optimized builds they are ignored and they just crash on all others cases.
better to handle these errors.

posix/JackPosixSemaphore.cpp
milloni: yep, will fix

milloni:
> the asserts against clocktime return value seem dangerous to me.
> on optimized builds they are ignored and they just crash on all others cases.
> better to handle these errors.

Ah, yes - I was planning to do something about this, and then forgot. The only ways `clock_gettime()` can fail are 1) programmer error, 2) requested clock not supported. It makes sense to handle 1) with an assertion since these should never happen in the program. For 2), my plan was to check if CLOCK_MONOTONIC is available *once*, and then, if it's not, forgo this check-for-skips dance altogether. I'll do that in the next revision of this PR.

milloni:
> I thought there was a non-portable call to setup a semaphore clock..
> (I mean, why did you not use it directly?)

Are you talking about http://sourceware-org.1504.n7.nabble.com/PATCH-Semaphores-add-sem-timedwait-monotonic-GNU-extension-BZ-14717-td550206.html#a550467 ?
This is a proposed patch to glibc but it's not merged or even ready yet.
 
> also note that posix-semaphore c++ code is not used under linux, that is why CI passes even though the code itself should not build.

Are you sure? On my Ubuntu 16.04 LTS:

```
$ ./waf configure
<snip>
$ ./waf build
<snip>
[ 29/215] cxx: posix/JackPosixSemaphore.cpp -> build/posix/JackPosixSemaphore.cpp.1.o
[ 30/215] cxx: posix/JackPosixProcessSync.cpp -> build/posix/JackPosixProcessSync.cpp.1.
<snip>
```

milloni:
Ok, I pushed the v2 revision. I think the CI is broken

milloni:
I tested this on a fork of 1.9.9.5. I tested legacy paths (i.e CLOCK_MONOTONIC not supported) by changing `if(res != 0)` to `if(true)` where necessary. I don't have a way of testing on MacOS/other Unix and I'm not really sure how I would go about testing against current master (it probably won't work on the system that I'm testing on).

7890:
The travis build fails because:
```
E: Unable to locate package gcc-6
E: Unable to locate package g++-6
E: Couldn't find any package by regex 'g++-6'
The command "./.ci/install-deps.sh" failed and exited with 100 during .
Your build has been stopped.
```
This is something unrelated to this PR.

milloni:
Hey I just remember this. Any chance you could accept it?

falkTX:
tbh I do not like this approach very much. the issue should be fixed elsewhere, jack is not the only software that has this problem.
also, we should use `CLOCK_MONOTONIC_RAW` if possible

milloni:
Sorry, I missed the last message   

@falkTX I think you're right, although glibc folks have already had this discussion and they're not going to fix it, because doing so would supposedely break posix API. They are working on a different API that doesn't have this problem but it's an extension (not in posix) and ultimately it's downstream software that will have to make the switch.   

Honestly I'm a bit surprised at this counter-intuitive behavior but apparently it's a "feature" not a bug

PR #420, opened 2019-01-25
README: Add repology repository badge

luzpaz:
This badge links to a chart that shows where jack2 is the package ecosystem. ~Apologies,~ I was ~un~successful formatting the .rst so the badges could be side-by-side.

Edit: See https://github.com/luzpaz/jack2/blob/README-repology/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,8 +2,10 @@
 ################################
 
 .. image:: https://travis-ci.org/jackaudio/jack2.svg?branch=master
-   :target: https://travis-ci.org/jackaudio/jack2
-
+   :target: https://travis-ci.org/jackaudio/jack2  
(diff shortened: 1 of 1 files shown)


falkTX:
This could be useful. Sad to see current stable Debian and Ubuntu still have an old version that does not have meta-data support.
Will merge now, thanks!

trebmuh:
Debian testing and unstable have 1.9.14 .

falkTX:
Indeed. Though those are not "stable" yet.

trebmuh:
Correct.
You know how Debian works, 1.9.14 wasn't there when Debian (current) stable got out. So it's not in there.
It'll then be on the next Debian stable.

PR #426, opened 2019-02-08
memops-alsa_bad_sound-fix-1

rncbc:
This s about to revert ALL of the following series of commits, as given by:
```
> git log --oneline  c5a0f5ea...831163e5^

c5a0f5ea oss_driver: Use float to S32 conversion if requested
148c8d8e alsa_in/out: Use float to S32 conversion if requested
d017e1ff alsa_driver: Use float to S32 conversion if requested
bb99e09b memops: Provide function for float to S32 conversion
b4ea23df memops: Align S24LE and S32LE to float conversion
244fc27e memops: Provide function for S32 to float conversion
4455fe02 memops: Align float to S24LE and S32LE conversion
a82f3f2f memops: Remove not used conversion macros
e7532543 memops: Use right-aligned S24LE to float conversion
831163e5 memops: Use right-aligned float to S24LE conversion

```
Apparently at least one of the above is causing severe sound anomalies eg. glitching/clipping distortion, and/or also the opposite effect of extreme low level signal and perceived constant drop outs.

You can either revert explicitly starting from the top and step down or merge this PR.

### DIFF
--- a/common/memops.c
+++ b/common/memops.c
@@ -73,8 +73,8 @@
    So, for now (October 2008) we use 2^(N-1)-1 as the scaling factor.
 */
 
-#define SAMPLE_24BIT_SCALING  8388607
-#define SAMPLE_16BIT_SCALING  32767
+#define SAMPLE_24BIT_SCALING  8388607.0f
+#define SAMPLE_16BIT_SCALING  32767.0f
 
 /* these are just values to use if the floating point value was out of range
    
@@ -414,10 +414,8 @@ void sample_move_d32u24_sS (char *dst, jack_default_audio_sample_t *src, unsigne
 
 void sample_move_dS_s32u24s (jack_default_audio_sample_t *dst, char *src, unsigned long nsamples, unsigned long src_skip)
 {
-	const jack_default_audio_sample_t scaling = 1.0 / (SAMPLE_24BIT_SCALING << 8);
-
 #if defined (__ARM_NEON__) || defined (__ARM_NEON)
-	float32x4_t factor = vdupq_n_f32(scaling);
+	float32x4_t factor = vdupq_n_f32(1.0 / SAMPLE_24BIT_SCALING);
 	unsigned long unrolled = nsamples / 4;
 	while (unrolled--) {
 		int32x4_t src128;
@@ -437,8 +435,7 @@ void sample_move_dS_s32u24s (jack_default_audio_sample_t *dst, char *src, unsign
 				break;
 		}
 		src128 = vreinterpretq_s32_u8(vrev32q_u8(vreinterpretq_u8_s32(src128)));
-		/* sign extension - left shift will be reverted by scaling */
-		int32x4_t shifted = vshlq_n_s32(src128, 8);
+		int32x4_t shifted = vshrq_n_s32(src128, 8);
 		float32x4_t as_float = vcvtq_f32_s32(shifted);
 		float32x4_t divided = vmulq_f32(as_float, factor);
 		vst1q_f32(dst, divided);
@@ -450,6 +447,8 @@ void sample_move_dS_s32u24s (jack_default_audio_sample_t *dst, char *src, unsign
 #endif
 
 	/* ALERT: signed sign-extension portability !!! */
+
+	const jack_default_audio_sample_t scaling = 1.0/SAMPLE_24BIT_SCALING;
 
 	while (nsamples--) {
 		int x;
@@ -470,8 +469,7 @@ void sample_move_dS_s32u24s (jack_default_audio_sample_t *dst, char *src, unsign
 		x <<= 8;
 		x |= (unsigned char)(src[0]);
 #endif
-		/* sign extension - left shift will be reverted by scaling */
-		*dst = (x << 8) * scaling;
+		*dst = (x >> 8) * scaling;
 		dst++;
 		src += src_skip;
 	}
@@ -479,11 +477,10 @@ void sample_move_dS_s32u24s (jack_default_audio_sample_t *dst, char *src, unsign
 
 void sample_move_dS_s32u24 (jack_default_audio_sample_t *dst, char *src, unsigned long nsamples, unsigned long src_skip)
 {
-	const jack_default_audio_sample_t scaling = 1.0 / (SAMPLE_24BIT_SCALING << 8);
-
 #if defined (__SSE2__) && !defined (__sun__)
 	unsigned long unrolled = nsamples / 4;
-	__m128 factor = _mm_set1_ps(scaling);
+	static float inv_sample_max_24bit = 1.0 / SAMPLE_24BIT_SCALING;
+	__m128 factor = _mm_set1_ps(inv_sample_max_24bit);
 	while (unrolled--)
 	{
 		int i1 = *((int *) src);
@@ -496,8 +493,7 @@ void sample_move_dS_s32u24 (jack_default_audio_sample_t *dst, char *src, unsigne
 		src+= src_skip;
 
 		__m128i src = _mm_set_epi32(i4, i3, i2, i1);
-		/* sign extension - left shift will be reverted by scaling */
-		__m128i shifted = _mm_slli_epi32(src, 8);
+		__m128i shifted = _mm_srai_epi32(src, 8);
 
 		__m128 as_float = _mm_cvtepi32_ps(shifted);
 		__m128 divided = _mm_mul_ps(as_float, factor);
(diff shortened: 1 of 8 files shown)


7890:
Most likely introduced in https://github.com/jackaudio/jack2/pull/397

rncbc:
@7890 : it shows up while using the normal alsa backend (and not quite using the alsa_in/out bridges) most evident on punchy energetic bass sounds start clipping/glitching/distorted badly; lowering  headroom helps but that's not the point here: it was sounding quite as right as before, as in jack1, for years. and reverting the changes above gets the system back to sane status.

cheers

7890:
@rncbc that's a severe effect. I'm all for sane status!
For the archives if possible some artifacts of working vs. faulty signals would be nice to have. Either as wave or imagery.

falkTX:
@rncbc thanks for the quick action.

@twischer-adit please have a look at what is going on here.

twischer-adit:
@rncbc Which sound card are you using? May be do you also know the name of the driver you are using? Could you provide the logging output of JACK? So that we could get an idea with which format you are opening the sound device.

rncbc:
@twischer-adit: as said on irc this was evident on two different machines and soundcards:

1) desktop, snd-ice1712;
2) laptop, snd-intel-hda;

both cases had the same clipping distortion issues (often like a signed-integer range pcm signal overflow);

once the series is reverted, all is back to previous sane and pristine sounding--even though the signal may go over 0dbfs while in internal graph processing, which is perfectly legal in jack, and way *before* hitting the alsa pcm back-end and the D/A borderline (ie. the driver back-end).

cheers

7890:
I can confirm the issue now with on-board intel PCM. It wasn't obvious but suddenly a session started to have clicks where there shouldn't be any. It was a jack installation with the bad memops. Thankfully this was detected and reverted!

twischer-adit:
> as said on irc

Hi all,
as far as I got there is no archive for the IRC channel available. Therefore I am not able to read the related discussion. It would be helpful if anyone could provide a reproduction scenario with simple playback. May be also attach the media file which produces the clicks.

It looks like that `snd-ice1712` supports `S32_LE`.
```
~/linux/sound/pci/ice1712$ grep -r S32
ice1724.c:	.formats =		SNDRV_PCM_FMTBIT_S32_LE,
ice1724.c:	.formats =		SNDRV_PCM_FMTBIT_S32_LE,
ice1724.c:	.formats =		SNDRV_PCM_FMTBIT_S32_LE,
ice1712.c:	.formats =		SNDRV_PCM_FMTBIT_S32_LE,
ice1712.c:	.formats =		SNDRV_PCM_FMTBIT_S32_LE,
```
Therefore I would expect JACK is also opening the sound device with `S32_LE`. Some `snd-intel-hda` seems also to support `S32_LE`:
```
~/linux/sound/pci/hda$ grep -r S32
hda_eld.c:				formats |= SNDRV_PCM_FMTBIT_S32_LE;
hda_eld.c:				formats |= SNDRV_PCM_FMTBIT_S32_LE;
patch_ca0132.c:				channels, SNDRV_PCM_FORMAT_S32_LE, 32, 0);
patch_hdmi.c:	(SNDRV_PCM_FMTBIT_S16_LE | SNDRV_PCM_FMTBIT_S32_LE)
```
@rncbc Was both of your cards are using `S32_LE`?

@7890 Could you tell which driver your sound card is using and which format the sound card was opened by JACK? Also `S32_LE`.
Unfortunately I have no sound card available which supports `S32_LE` only `S24_LE`. I will try to reproduce the issue with `snd-aloop` or `snd-dummy`

falkTX:
IRC logs:
```
16:19 < rncbc> ouch! houste i think we have a severe problem on jack2-git_head (a752bfa)... sound is distorted and clipping badly
16:19 < rncbc> houston*
16:22 < rncbc> reverted to (my) previous sane build (c5a16d0) and all is back to normal
16:31 < rncbc> i guess some1 has to bisect this and find the culprit
16:34 < rncbc> summary: current master HEAD a752bfa is bad; c5a16d0 is good.
16:39 < dvzrv> damn
17:02 < rncbc> after git bisect : 831163e51640aad2fbda7710c44bc406f167df85 is the first bad commit
17:03 < rncbc> commit 831163e51640aad2fbda7710c44bc406f167df85
17:03 < rncbc> Author: Timo Wischer <twischer@de.adit-jv.com>
17:03 < rncbc> Date:   Fri Nov 16 09:26:05 2018 +0100
17:03 < rncbc>     memops: Use right-aligned float to S24LE conversion
17:03 < rncbc>     
17:03 < rncbc>     ALSA expects right-aligned samples (0x00******) as mentioned in the
17:03 < rncbc>     source code for SND_PCM_FORMAT_S24_LE:
17:03 < rncbc>     Signed 24 bit Little Endian using low three bytes in 32-bit word
17:03 < rncbc>     See http://git.alsa-project.org/?p=alsa-
17:03 < rncbc>     lib.git;a=blob;f=include/pcm.h;h=5b0782315585de1d5ab82c9f2036b62c168f5a48;hb=HEAD#l140
17:03 < rncbc>     
17:03 < rncbc>     Signed-off-by: Timo Wischer <twischer@de.adit-jv.com>
17:08 -!- wargreen_ [~wargreen_@37.164.119.219] has quit [Ping timeout: 244 seconds]
17:23 < rncbc> but there's more
17:25 < rncbc> only reverting all changes made to common/memops.* example-clients/alsa_*.c and of course linux/alsa/alsa_driver.c, things get back to normal (sonic wise)
17:26 < rncbc> so the above 831163e is only the first of a series that broke jack2 git!
17:47 < dvzrv> falktx: probably time to do some reverts
17:51 < rncbc> all git log c5a0f5ea...831163e5^ are in to revert
17:54 <@falktx> dvzrv: I need to leave, do you mind doing that?
17:55 <@falktx> or it can wait till tomorrow I guess
18:00 < dvzrv> yeah, think so
18:00 < rncbc> i'll make a PR for just that
18:14 < rncbc> done: https://github.com/jackaudio/jack2/pull/426
22:24 <@falktx> rncbc: thanks, just applied
```

7890:
```
card 0: MID [HDA Intel MID], device 0: ALC269 Analog [ALC269 Analog]
  Subdevices: 0/1

$ cat /proc/asound/card0/pcm0c/sub0/hw_params 
access: MMAP_INTERLEAVED
format: S32_LE
subformat: STD
channels: 2
rate: 44100 (44100/1)
period_size: 512
buffer_size: 1536
```
My observation is that with 'bad memops' in place, every value > 0dBFS will crackle. This can be a bad cut, an inter-sample peak or just a value slightly above 0. I can easily test this with a player, using "normal" jackd, i can go a bit over 0dBFS without hearing artifacts until hitting around +3dBFS. With 'bad memops' there is crackling from the first sample above 0. One could say it's normal that >0 will clip however it did not do it before in an audible way. That's how it behaves.
PS: I started a new session with stems and thought it's crazy how many crackles are introduced by obviously bad cuts. After starting to fix a few of them it came to mind that it could be jack. After updating jack all was "normal" again. I can't tell why values above 0 don't clip the same way, it could be that there is some "hidden" headroom.

twischer-adit:
Hi all,

without my modification `sample_move_d32u24_*()` is used for the conversion from float to S32_LE. This function uses `float_24u32` (see https://github.com/jackaudio/jack2/blob/master/common/memops.c#L411).

`float_24u32` converts `1.0f` into `0x7FFFFF00` (see https://github.com/jackaudio/jack2/blob/master/common/memops.c#L137 and https://github.com/jackaudio/jack2/blob/master/common/memops.c#L76) but a S32_LE sound card supports samples up to `0x7FFFFFFF`. Therefore there is a gap of 255 (=0xFF). This results in a maximal float value of 1.000000119f (=1.0f * 0x7FFFFFFF / 0x7FFFFF00). Therefore there is currently some headroom (but not up to +3dB aka 1.413f).

A S16_LE sound card uses `sample_move_d16_*()` which uses `float_16()`. This function converts `1.0f` into 0x7FFF. Therefore there is no headroom. That means S32_LE behaves currently different as S16_LE conversion.

@7890 Do you really have only cracks with values >+3dB? I would already expect cracks at >+0.000001034dB (~1.000000119f)

7890:
I was surprised too. No obvious audible artifacts until going really high above 0. The other interesting thing is that the effect is different: while normal jack starts to distort (it sounds pretty analog) for say +6dBFS, with bad mempos the crackle is much more distinctive and starts right from >0. My feeling is that there is something else going on, it can't be just the difference of +0.000001034dB.

PR #428, opened 2019-02-09
templated systemd user unit

dvzrv:
This adds a templated systemd user unit (`jack@.service`), which relies on a configuration profile (either below `/etc/jack/` or `$XDG_CONFIG_HOME`).
An example configuration profile is also provided.
This is a static approach to starting JACK (in comparison to starting it via dbus), but is very useful in a context, where other JACK aware software (also started via the systemd user unit approach, e.g. on a headless system) should rely on JACK (with a specific configuration) being started.
Additionally, it has the upside of the JACK log ending up in the system journal.

I'm unsure, whether this should be installed via the build system, as there are Linux distributions out there, that don't use systemd (an optional flag wouldn't hurt I guess).

In any case: I have been using this approach for many years now and think it's quite useful.


dvzrv:
Getting back to this: Waf is pretty bizarre and I don't really know how to hook up installation of the files into `wscript`. Any hints?
Seems pkgconfig exposes the place where systemd user units should be installed on a given system via `pkg-config systemd --variable=systemduserunitdir` (e.g. `/usr/lib/systemd/user`).

dvzrv:
I switched to xenial for all builds, because trusty just started breaking builds by not finding `sd-daemon.h`. Also, it's basically dead and we should not even build against it. ;-)

The unit and configuration seem to install fine to the defined locations.

dvzrv:
@karllinden I'm a bit lost here. Can you maybe help in finding a way to only optionally (by specifying `--systemd-unit` during configure) install the unit files?
It seems, that `Options.options.systemd_unit` is never picked up and thus the files never installe.d

7890:
I wonder if not finding 'sd-daemon.h' warrants a switch of the build system, even if it's just travis..
Doing this on 14.04 shows there is that header:
```
$ apt-file search sd-daemon.h
libsystemd-daemon-dev: /usr/include/systemd/sd-daemon.h
```

dvzrv:
@7890 might be due to weird packages from the included ppa.
In the end it doesn't really make sense to support trusty in the first place I guess.

Which CI system are you aiming at? Most of the install/build stuff could probably easily be reused.

7890:
I'm just not seeing that changing the build host release version to achieve "templated systemd user unit" is reasonable at this point. It's an indication that jack can compile on older systems too and chances are good that it will compile nicely on any follow-up release of that.

dvzrv:
@7890 Maybe you missed a point in my message on that topic:
My change to xenial has nothing to do with my pull request directly, but (most likely) the one that introduced the systemd notify feature for jackd (which in turn relies on a library it can't find all of a sudden, which in turn is a problem of the targetted OS).
AFAICS, [waf couldn't find the systemd library](https://travis-ci.org/jackaudio/jack2/jobs/539340626#L896) (for environmental/integrational reasons, [not because of this pull request](https://travis-ci.org/jackaudio/jack2/jobs/539340626#L905)), [but tried to build against it nonetheless](https://travis-ci.org/jackaudio/jack2/jobs/539340626#L1421) (which is a bug in the waf setup).
As I have no further means to track down *why* something in a travis black box VM is not available all of a sudden (besides very time consuming trial and error), the easiest and also most logical way to fix this, is to switch to an operating system version, that is actually *supported by its own upstream* (when one realizes, that this is not the case anymore). The success in doing so proves, that the issue was indeed an underlying problem in the trusty setup (or VM, or ppa, etc.).

That being said: [Ubuntu trusty has been EOL for more than a month now](https://en.wikipedia.org/wiki/Ubuntu_version_history#Ubuntu_14.04_LTS_(Trusty_Tahr)). There is absolutely no reason to either support it (or whatever hacks have to be applied on top of it now to make it work for travis) or build against it. From this month onwards, there will be nothing reproducible or standard about the travis setup for trusty VMs.
Therefore, I'm not surprised, builds just start breaking now, but I'm surprised there's even a discussion about changing to Xenial...

> and chances are good that it will compile nicely on any follow-up release of that.

From a build environment point of view, this is wishful thinking. Ubuntu definitely introduces environmental changes to their packaging eco-system, as we've just witnessed (e.g. realizing, that it makes more sense to have `libsystemd-dev`, instead of `libsystemd-journal-dev`, `libsystemd-login-dev`, `libsystemd-id128-dev`, `libsystemd-daemon-dev` and more).
My stance is, that the jack2 project is (time wise) better off not maintaining someone else's unsupported OS hacks and instead ensuring, that things build on *operating system versions, that their upstream supports*!

However, if there are supposed to be tests against trusty (which I really not recommend, as it means dealing with the hacks of a third party system - travis), then there should be a separate build stage targetting it (that's allowed to fail, as there might be more environmental changes to the trusty images/ package sources in the future), but it should by no means be used in the stages deciding whether a build fails or not!

dvzrv:
@karllinden Sorry for the ping. The problem resolved itself, after not being stupid and actually supplying the right flags to `./waf configure` :>

dvzrv:
I have now also templated the use of `PREFIX` in the service file, so it can be used in other scenarios (e.g. `/bin`, `/usr/local/bin`).

7890:
@dvzrv  .. not convinced. From the offered OS' on Travis
```
    Ubuntu Xenial 16.04
    Ubuntu Trusty 14.04 default
    Ubuntu Precise 12.04
```
I'd even tend to go downwards to 12.04... 
Just because OS is EOL doesn't mean it won't work anymore. A static (no more updates) build host can have benefits too. We'd want the source to be in shape for compilation on such hosts. It should be a no-brainer to compile JACK on any of these platforms.

IIUC the need for sd-damon.h was effectively introduced in 9fe6486c6e289c24d4d7d264ef49ff3fa64eeb7d
I don't understand why waf can't find the library, that's strange indeed.

What I dislike in this PR is that it's mixing the build host switch with something that could be done (?) with a simple (Linux only) shell script that creates the necessary systemd files from available environment infos, eventually put to a `contrib/systemd` directory. Its eventually a bit late to discuss this..

@dvzrv please share an instance of the systemd related installables / files that come out of the waf process, I'd like to get an idea of the complexity.

7890:
@dvzrv looking at the template file:
https://github.com/jackaudio/jack2/blob/2835c4041016d4fb8a849fe7bdef90956de9d8e7/systemd/jack%40.service.in
thinking out loud it can easily be filled with a script that handles all variables nicely as command line options. It can be verbose and even have a --help option if needed. If there's no hard dependency on anything going on in waf, it's not a loss to have it as a separate unit.
For the place to store the template and eventually the script, having a `systemd` directory in the root of the project worries me. It could go to `linux/systemd` or in that particular case (no code to compile involved, Linux only, optional utility) as proposed to `contrib/systemd/*`. A contrib folder would indicate that it contains useful tools for particular interests with no dependency on the build step and no code that is nested with the code base, and that tools therein are not officially supported in case of (tangent) issues with starting jackd via systemd.

7890:
Some inspiration for a .sh script, **incomplete**
```
#!/bin/sh

#environment checks (systemd prerequisites, ..)
#argument parsing & validation (if needs any)...
#default values ...

TMPFILE="`mktemp`"
JACKD_BINARY="`which jackd`"
i="foo" #

cat - > "$TMPFILE" <<  __EOF__
[Unit]
Description=JACK server using ${i}.conf profile
Documentation=man:jackd(1)
After=sound.target local-fs.target

[Service]
Type=notify
EnvironmentFile=-/etc/jack/${i}.conf
EnvironmentFile=-${HOME}/.config/jack/${i}.conf
ExecStart=${JACKD_BINARY} \$JACK_OPTIONS -d \$DRIVER -d \$DEVICE \$DRIVER_SETTINGS
LimitRTPRIO=95
LimitRTTIME=infinity
LimitMEMLOCK=infinity
# Caution: use on memory-limited devices only
# OOMScoreAdjust=-1000

[Install]
WantedBy=default.target
__EOF__
cat "$TMPFILE"

#optionally let user reviev output ...

#store to xxyy
echo "# $XDG_CONFIG_HOME"

#success?

#clean up ...
#rm -f "$TMPFILE"

```

7890:
@dvzrv as for the missing header, I don't understand fully what's going on.
In this build of recent development branch:
https://travis-ci.org/jackaudio/jack2/jobs/538125696
It shows 
```
Checking for header systemd/sd-daemon.h  : yes 
Checking for library systemd             : not found 
...
Use systemd notify                          : no
```
and the build won't show undefined reference. Are you saying that when --systemd is given AND library was not found there will be the build error? That would indeed be something to solve in the waf file.

dvzrv:
@7890 I have spent days on testing a solution that is *integrated with the build system*.
Presenting an arbitrary, separate and completely *unintegrated* solution to me (that is on top *not conform* to anything else in this repository, e.g. the dbus service integration), *after* doing all this work honestly just feels like mockery by now. There is absolutely *no upside* to having yet another script in here, that does the exact same thing as the build system (but with more code!).

I don't understand the issue with wherever in this repository the files are present either, after ensuring my support for this "feature" (also for the future). It literally doesn't make any difference and to me feels like changing something for change's sake.
Sure, I can move the files below `linux/systemd`, but does the `dbus` directory also have to go below `linux` now? Would you also create a separate script for installing the `dbus` service file?
After all, this pull request is about installing two plaintext files to the correct location.

A note on Travis: Just because they offer legacy systems, doesn't mean that that is right, or that you have to use them (the maintenance burden is yet another issue that is added on top). You also wouldn't start building against Windows XP or macOS 10.8, just because you can.

falkTX:
@7890 just because I am not as active as you wish I'd be, it does not give you the right to act like this.
I have been seeing this PR together with @dvzrv for a long time, even talking about and reviewing face to face.
You are not the jack maintainer, you do not get to make the decisions of what we support or not.

This was pretty much ready to merge, I only asked @dvzrv to do a "small" change, which he very kindly did.

Regarding CI OS support. I agree it makes no sense to support an EOL version.
Even though I still use 14.04 daily, I understand it is an EOL system that needs to be phased out so developers and packagers can focus on newer things.

PR #434, opened 2019-02-24
Solving problems while compiling jack2 on macOS X with dbus support

shurizzle:
I have a problem while compiling jack2 on macOS X with dbus support because sysinfo.h is a linux exclusive.
I used sysctl on macOS to get uptime and it just works.

### DIFF
--- a/dbus/controller.c
+++ b/dbus/controller.c
@@ -27,13 +27,13 @@
 #include <dbus/dbus.h>
 #include <assert.h>
 #include <unistd.h>
-#include <sys/sysinfo.h>
 #include <errno.h>
 
 #include "controller.h"
 #include "controller_internal.h"
 #include "xml.h"
 #include "reserve.h"
+#include "uptime.h"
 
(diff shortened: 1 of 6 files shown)


dbus/controller.c
```diff
@@ -787,18 +787,18 @@ void
...
 
     if (controller_ptr->pending_save == 0)
     {
         return;
     }
 
-    if (sysinfo(&si) != 0)
+    if ((ut = uptime()) < 0)
     {
         jack_error("sysinfo() failed with %d", errno);
```
falkTX: you forgot UPTIME_FUNCTION_NAME here

falkTX:
looks good, though I wonder about jack checking for errno in case of failure.
Does sysctl set it too?

shurizzle:
I'm sorry I forgot to push it.
Yes, sysctl sets errno :)

falkTX:
Alright, thank you!
This is clean and simple enough.

PR #435, opened 2019-02-26
Jack retry after poll timeout

DeviLaxmii:
On ALSA poll time out, retry polling after executing the recovery

### DIFF
--- a/linux/alsa/alsa_driver.c
+++ b/linux/alsa/alsa_driver.c
@@ -1370,6 +1370,14 @@ alsa_driver_wait (alsa_driver_t *driver, int extra_fd, int *status, float
 
 		poll_ret = jack_get_microseconds ();
 
+		if (poll_result == 0) {
+			jack_error ("ALSA: poll time out, polled for %" PRIu64
+				    " usecs",
+				    poll_ret - poll_enter);
+			*status = -5;
+			return 0;
+		}
+
         // JACK2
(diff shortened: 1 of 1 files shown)


twischer-adit:

kmatheussen:
Looks good. Just wonder, what is the advantage of retrying instead of setting a higher timeout?

DeviLaxmii:
> 
> 
> Looks good. Just wonder, what is the advantage of retrying instead of setting a higher timeout?
Hi @kmatheussen ,

Actually increasing the timeout did not help the situation. I tried retry on poll for 5 times(Without the XRUN recovery), with the poll timeout as 3Sec each, So totally waited for 15Secs, But poll still timed out.
The sound card could not respond even after 15secs, but recovery healed the issue.
This patch could be helpful for other users with similar buggy sound cards.

Thank you for reviewing!!

kmatheussen:
> This patch could be helpful for other users with similar buggy sound cards.

Great! This is also an issue on Windows. For some reason, you often have to press the start button twice in qjackctl on that platform. It would be a significant improvement of jack if this patch fixes that issue.

kmatheussen:
> It would be a significant improvement of jack if this patch fixes that issue.

And,,, I just realized that this patch was for the alsa driver. :-)
But maybe the same thing can be done to the portaudio driver.

PR #436, opened 2019-02-27
Import metadata keys from jackey, cleanup and reorder definitions

falkTX:
Opening PR to leave space for discussion.
Keys were imported from https://github.com/drobilla/jackey project, leaving the URL exactly the same (they were using jackaudio.org) but now officially supported.

We need to write some description for the keys that were already there.

@drobilla since you did the new ones and are quite skilled on wording (being native english speaker), do you mind trying to write a description for the "old" jack metadata keys? (see the "..." header comments on this PR)
Thanks in advance!

### DIFF
--- a/common/jack/metadata.h
+++ b/common/jack/metadata.h
@@ -40,6 +40,15 @@ extern "C" {
 
 /**
  * A single property (key:value pair).
+ *
+ * Although there is no semantics imposed on metadata keys and values, it is
+ * much less useful to use it to associate highly structured data with a port
+ * (or client), since this then implies the need for some (presumably
+ * library-based) code to parse the structure and be able to use it.
+ *
+ * The real goal of the metadata API is to be able to tag ports (and clients)
+ * with small amounts of data that is outside of the core JACK API but
+ * nevertheless useful.
  */
 typedef struct {
     /** The key of this property (URI string). */
@@ -205,7 +214,8 @@ int jack_set_property_change_callback (jack_client_t*             client,
                                        void*                      arg);
 
 /**
- * ...
+ * A value that identifies what the hardware port is connected to (an external
+ * device of some kind). Possible values might be "E-Piano" or "Master 2 Track".
  */
 extern const char* JACK_METADATA_CONNECTED;
 
@@ -223,12 +233,16 @@ extern const char* JACK_METADATA_CONNECTED;
 extern const char* JACK_METADATA_EVENT_TYPES;
 
 /**
- * ...
+ * A value that should be shown when attempting to identify the
+ * specific hardware outputs of a client. Typical values might be
+ * "ADAT1", "S/PDIF L" or "MADI 43".
  */
 extern const char* JACK_METADATA_HARDWARE;
 
 /**
- * ...
+ * A value with a MIME type of "image/png;base64" that is an encoding of an
+ * NxN (with 32 < N <= 128) image to be used when displaying a visual
+ * representation of that client or port.
  */
 extern const char* JACK_METADATA_ICON_LARGE;
 
@@ -244,7 +258,9 @@ extern const char* JACK_METADATA_ICON_LARGE;
 extern const char* JACK_METADATA_ICON_NAME;
 
 /**
- * ...
+ * A value with a MIME type of "image/png;base64" that is an encoding of an
+ * NxN (with N <=32) image to be used when displaying a visual representation
+ * of that client or port.
  */
 extern const char* JACK_METADATA_ICON_SMALL;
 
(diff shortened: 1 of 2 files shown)


falkTX:
From IRC:
```
17:37 <+rgareus> falktx: "PRETTY_NAME" is a human readable name to be used instead of the port-name in GUIs
17:38 <+rgareus> falktx: icon small/large -- I think qjackctl uses those. I don't know what the semantics are. probably a file-name to an image
17:39 <+las> falktx: where is "connected" documented/mentioned?
17:41 <+las> hmm, the URIs are also supposed to be valid URLs
17:41 <+las> falktx: i have no memory of what "connected" was for. i also know of nothing uses it
17:43 <+rgareus> las: they are valid URLs, or you do mean that http://jackaudio.org/metadata/pretty-name should be a html page with doc?
17:43 <+las> rgareus: tha latter
17:43 <+las> rgareus: s/valid/functioning/
17:44 <+las> rgareus: i have a memory of writing the page for pretty-name
18:08 <+las> rgareus: i did. it was a drupal node from Ye Olde Jackaudio.orge Web Site
```

7890:
On the page http://jackaudio.org/metadata there could be more or less a copy of 
https://github.com/drobilla/jackey/blob/master/jackey.h
To describe the available event-types etc.

drobilla:
@falkTX I could give it a shot, but I don't know what some of these were supposed to mean, either.

SpotlightKid:
I resurrected the `metadata` page from the Drupal site fom the wayback machine in the wiki:

https://github.com/jackaudio/jackaudio.github.com/wiki/JACK-Metadata-API

There doesn't seem to be a `metadat/pretty-name` page in the archive, though, but I havn't looked at all the archive dates yet.

SpotlightKid:
Also, when documenting metadata, please don't forget to *always* specify string encoding (if any) explicitly. Not only for `data`, but also for `key` and `type`.

Fo example, is it correct to assume that `key` is always ASCII encoded? Is puny-code allowed?

7890:
Excellent, thank you @SpotlightKid !

We need to make sure that the text is still 100% accurate and reflecting what's available.

As for the encoding of strings, this is something eventually not limited to metadata key/values.
The basic rule is that ASCII will always work. Let's take an example where the text is encoded in UTF-8. If it doesn't need to use more than one byte per character, it's still ASCII so we're fine. If it contains characters with multiple bytes and still fits into the buffer, it will depend on the processes that read the buffer and how they eventually "print" it. If this is done within an UTF-8 aware context, everything is still fine. One odd case is when the buffer limit cuts a multibyte character. There are no checks in place to parse strings to detect such cases, it's handled as a char buffer expected to be ASCII now, as far as i can see. This is a good point - it should be clearly mentioned in the documentation.

7890:
One prime use case for metadata keys is to flag MIDI ports to indicate the data format/type pushed via this non-fixed-length channel. Patchage does support this since a long time - it handles ports with an OSC key differently.
@drobilla @falkTX How do you think we can create a foundation for custom keys for jack that are still organized in some way, I'm thinking of LV2 which seems to have created an ecosystem with a well-defined way to extend a basic set of standard-provided keys with new keys. What's your thoughts about this, what should we do.

common/jack/metadata.h
```diff
@@ -204,19 +204,88 @@ int jack_set_property_change_callback (jack_client_t*             client,
...
+/**
+ * ...
+ */
+extern const char* JACK_METADATA_ICON_SMALL;
+
+/**
+ * Order for a port.
+ *
+ * This is used to specify the best order to show ports in user interfaces.
+ * The value MUST be an integer.  There are no other requirements, so there may
```
SpotlightKid: A common practice with clients seems to be to set the property type to `http://www.w3.org/2001/XMLSchema#integer`. If this is a requirements or encouraged, it should be mentioned here.

Also, the value is encoded as a decimal numeric string, not as binary.

common/jack/metadata.h
```diff
@@ -204,19 +204,88 @@ int jack_set_property_change_callback (jack_client_t*             client,
...
+ * be gaps in the orders for several ports.  Applications should compare the
+ * orders of ports to determine their relative order, but must not assign any
+ * other relevance to order values.
+ */
+extern const char* JACK_METADATA_ORDER;
+
+/**
+ * ...
+ */
+extern const char* JACK_METADATA_PRETTY_NAME;
```
SpotlightKid: A common practice with clients seems to be to set the property type to `text/plain`. If this is a requirements or encouraged, it should be mentioned here.

It should also be specified whether non-ASCII chars are allowed in the property value and how they should be encoded (UTF-8? Or would an encoding have to be specified in the property type (e.g. like in a HTTP `Content-Type` header?)

SpotlightKid:
I just stumbled over this old thread from the 'jack-devel' mailing list:.

7890:
archive check 1 2
https://jackaudio.github.io/mailing_list_archive/mail/1380636803.24393_0.ltw:2,/index.html
:1234:

common/jack/metadata.h
falkTX: The docs on the `type` say:
> If type is NULL or empty, the data is assumed to be a UTF-8 encoded string (text/plain)

I think that is clear enough, the values are utf8 strings by default

PR #440, opened 2019-03-17
Fix client UUID gen, and some warnings

falkTX:


falkTX:
Testing is appreciated

7890:
I'm using jack_property to test and it still has issues with client names.

Some warnings while compiling `fix-client-uuid-gen` branch (not sure relevant or if directly related to PR but to IDs):

```
../common/JackAPI.cpp: In function ‘char* jack_client_get_uuid(jack_client_t*)’:
../common/JackAPI.cpp:1986:86: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘jack_uuid_t {aka long unsigned int}’ [-Wformat=]
         snprintf(retval, sizeof(retval), "%d", client->GetClientControl()->fSessionID);
                                                                                      ^
../common/JackClient.cpp: In member function ‘virtual int Jack::JackClient::ClientNotify(int, const char*, int, int, const char*, int, int)’:
../common/JackClient.cpp:312:94: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘jack_uuid_t {aka long unsigned int}’ [-Wformat=]
                     snprintf(uuid_buf, sizeof(uuid_buf), "%d", GetClientControl()->fSessionID);
```

falkTX:
Should be working now, tested locally with Carla and a bunch of meta-data.

7890:
Best metadata support in jack2 so far +++

PR #447, opened 2019-03-26
Correct GPL licence to LGPL for files needed to build libjack (3)

sdrik:


falkTX:
Thanks for the quick PR, though would be nice to target develop branch for new stuff.

PR #448, opened 2019-03-26
Remove GPL dependency from JackTools.

vikwil:
JackTools is a part of the LGPL licenced jacklib and should not pull in
GPL licensed code.


falkTX:
Thanks

PR #449, opened 2019-03-26
Fix memory leak of parameter constraints.

vikwil:
A simple example calling jackctl_server_create, open, start, stop, close and finally destroy will leak memory allocated for parameter constraints. 

With this fix asan does not report any leaks when loading the "jack_dummy.so" driver.


7890:
Thanks @vikwil, merging to develop

PR #453, opened 2019-04-08
[metadata] Fix property key length to notification/callback requests.

rncbc:
Increase property key field length, from JACK_UUID_STRING_SIZE (37) to MAX_PATH+1 (256), as long to fit most metadata keys (URI) on change notification/callback requests.

### DIFF
--- a/common/JackRequest.h
+++ b/common/JackRequest.h
@@ -1632,7 +1632,7 @@ struct JackClientHasSessionCallbackRequest : public JackRequest
 struct JackPropertyChangeNotifyRequest : public JackRequest
 {
     jack_uuid_t fSubject;
-    char fKey[JACK_UUID_STRING_SIZE];
+    char fKey[MAX_PATH+1];
     jack_property_change_t fChange;
 
     JackPropertyChangeNotifyRequest() : fChange((jack_property_change_t)0)


7890:
@rncbc thanks for pointing this out. Metadata keys can be longer than 37 indeed.
I wondered if using MAX_PATH is reasonable in this case. It seems to relate to OS path length e.g. for loading dynamic libraries. Would it be worth introducing a new constant that is made especially to limit metadata key lengths?

rncbc:
it only makes for enough space for the notification request serialization; it doesn't put a limit in the metadata key length whatsoever, at least from the API point of view, which has no defined limit though.

SpotlightKid:
Just ran into this, when [my JacKPropertyChangeCallback](https://github.com/SpotlightKid/jack-matchmaker/blob/master/jackmatchmaker/__init__.py#L138) got called with a truncated property key string:

    DEBUG: Property 'http://jackaudio.org/metadata/pretty' on subject '4294967340' changed

which, of course, should have been 'http://jackaudio.org/metadata/pretty-name'.

7890:
It's a bug.. I merge this now - not sure about the constant name.
Thanks for PR and confirmation of bug!

PR #458, opened 2019-04-25
DOC: Update docs for jack_get_all_properties()

mgeier:
AFAICT, that's the actual behavior.

Same change for JACK1: https://github.com/jackaudio/headers/pull/8


falkTX:
yeah, that makes sense, thanks!

PR #461, opened 2019-05-14
Make jackdbus logging configurable and add some log hints

dpzmick:
Was looking for something to help out with; here's some small patches.

let me know what you think, figured it would be a nice principle-of-least-surprise feature to expand homedirs when simple to do so.

tested with things like

```
$ ./waf && XDG_CONFIG_HOME='~/' JACK_LOG_DIR='~/jacklogs/' build/dbus/jackdbus auto    
Waf: Entering directory `/home/dpzmick/programming/jack2/build'
Waf: Leaving directory `/home/dpzmick/programming/jack2/build'
'build' finished successfully (0.255s)
Directory "/home/dpzmick/jacklogs/" does not exist. Creating...
Directory "/home/dpzmick//jack" does not exist. Creating...
Directory "/home/dpzmick/jacklogs//jack" does not exist. Creating...
```

### DIFF
--- a/example-clients/jack_control
+++ b/example-clients/jack_control
@@ -157,17 +157,16 @@ def main():
         sys.exit(0)
 
     bus = dbus.SessionBus()
+    try:
+        controller = bus.get_object(service_name, "/org/jackaudio/Controller")
+        control_iface = dbus.Interface(controller, control_interface_name)
+        configure_iface = dbus.Interface(controller, configure_interface_name)
 
-    controller = bus.get_object(service_name, "/org/jackaudio/Controller")
-    control_iface = dbus.Interface(controller, control_interface_name)
-    configure_iface = dbus.Interface(controller, configure_interface_name)
-
-    # check arguments
-    index = 1
-    while index < len(sys.argv):
-        arg = sys.argv[index]
-        index += 1
-        try:
+        # check arguments
+        index = 1
+        while index < len(sys.argv):
+            arg = sys.argv[index]
+            index += 1
             if arg == "exit":
                 print("--- exit")
                 control_iface.Exit()
@@ -393,8 +392,11 @@ def main():
                 result = control_iface.RemoveSlaveDriver(name)
             else:
                 print("Unknown command '%s'" % arg)
-        except dbus.DBusException as e:
-            print("DBus exception: %s" % str(e))
+    except dbus.DBusException as e:
+        print("DBus exception: %s" % str(e))
+        print("Please check the dbus logs, most likely in the systemd journal (view with journalctl)")
+        print("If that doesn't help, you can try running `jackdbus auto` to check for errors in stderr")
+        sys.exit(1)
 
 if __name__ == '__main__':
     main()
--- a/example-clients/jack_control
+++ b/example-clients/jack_control
@@ -157,16 +157,23 @@ def main():
         sys.exit(0)
 
     bus = dbus.SessionBus()
+
     try:
         controller = bus.get_object(service_name, "/org/jackaudio/Controller")
         control_iface = dbus.Interface(controller, control_interface_name)
         configure_iface = dbus.Interface(controller, configure_interface_name)
+    except dbus.DBusException as e:
+        print("DBus exception: %s" % str(e))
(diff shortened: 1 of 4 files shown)


falkTX:
Kinda makes sense to me, but indentation on the lines you added is all over the place... you are mixing tabs and spaces.
Please fix this

example-clients/jack_control
```diff
@@ -157,17 +157,16 @@ def main():
...
-    # check arguments
-    index = 1
-    while index < len(sys.argv):
-        arg = sys.argv[index]
-        index += 1
-        try:
+    try:
+        controller = bus.get_object(service_name, "/org/jackaudio/Controller")
+        control_iface = dbus.Interface(controller, control_interface_name)
+        configure_iface = dbus.Interface(controller, configure_interface_name)
```
falkTX: I think it is better to separate into 2 `try` blocks here.
1 for setting up dbus objects and interfaces, the other as it was before.

example-clients/jack_control
```diff
@@ -393,8 +392,11 @@ def main():
                 result = control_iface.RemoveSlaveDriver(name)
             else:
                 print("Unknown command '%s'" % arg)
-        except dbus.DBusException as e:
-            print("DBus exception: %s" % str(e))
+    except dbus.DBusException as e:
+        print("DBus exception: %s" % str(e))
+        print("Please check the dbus logs, most likely in the systemd journal (view with journalctl)")
```
falkTX: jackdbus does not output to systemd journal by default, so how is this "most likely" ?

dpzmick:
@falkTX the file started out a mess, couldn't even figure out what convention to follow for tabs and space (many lines indented with a mix of the two, many functions inconsistent). What do you want for the new code?

example-clients/jack_control
dpzmick: jack install itself as a dbus activatable service

```
# process org.jackaudio.service.in -> org.jackaudio.service   
obj = bld(                                                    
        features = 'subst',                                   
        source = 'org.jackaudio.service.in',                  
        target = 'org.jackaudio.service',                     
        install_path = '${DBUS_SERVICES_DIR}/',               
        BINDIR = bld.env['PREFIX'] + '/bin')                  
```

and it looks like the dbus logs go to systemd journal on most distros, by default, so when the activation fails, the stderr for the jackdbus application _probably_ ends up in the journal. I'm not sure there's a way to do any better than that; as far as I can tell, dbus doesn't really have a well defined place to redirect stderr for dbus-activated services

7890:
Two things come to mind, could the expanduser be useful beyond this exact case? It could be at a place for other code to be available.

I wondered if the expansion can be made simpler, a quick search shows me this snippet:
```
#include <stdio.h>
#include <wordexp.h>

int main(int argc, char* argv[]) {
    wordexp_t exp_result;
    wordexp(argv[1], &exp_result, 0);
    printf("%s\n", exp_result.we_wordv[0]);
}

//  ./wordexp ~
```
https://stackoverflow.com/questions/15991241/how-do-i-achieve-tilde-expansion-in-c

dpzmick:
@7890 `wordexp` does a bunch of other things that are probably undesirable, e.g. passing it "~/$0/", it will expand the $0 as well.

example-clients/jack_control
dpzmick: fixed

7890:
My humble opinion here is that we either do a full `sh` equivalent expansion (using appropriate functions) or then rely on a fully qualified path with no shell variables in which case '~/' wouldn't be valid.

dpzmick:
I'm cool with that. I'll take the expand_user code out when I get a chance.

7890:
What are your thoughts @falkTX ?

dpzmick:
Thinking through it a bit more, the only time I'd probably ever stick a "~/" somewhere is in debugging, when I'm doing an override on a command line anyway. In that case it's probably a lot more sane to leave the string unquoted (or double quoted) and let the shell do it.

I'm more than happy to wait for additional input but I'm also more than happy to just rip the code out.

PR #476, opened 2019-06-09
Import metadata keys from jackey, cleanup and reorder definitions

darbyselector:
Signed-off-by: falkTX <falktx@falktx.com>

good work...  jack that body


darbyselector:
good lookin out

PR #482, opened 2019-07-01
alsa_driver: Exiting loop for need_playback and need_capture on XRUN …

DeviLaxmii:
…to execute recovery

Currently even on XRUN, the xrun recovery is not executed as the
need_playback or need_capture flag is still true and so it goes
and waits on poll again.

Signed-off-by: Laxmi Devi <Laxmi.Devi@in.bosch.com>


twischer-adit:

falkTX:
Looks good to me, merging now, thanks.

PR #493, opened 2019-08-26
jack2: possible race question

amiartus-adit-jv:
slave read and write should be done outside of graph processing to avoid race
conditions

Signed-off-by: Adam Miartus <amiartus@de.adit-jv.com>

### DIFF
--- a/common/JackAudioDriver.cpp
+++ b/common/JackAudioDriver.cpp
@@ -322,21 +322,21 @@ void JackAudioDriver::ProcessGraphSyncMaster()
     // fBeginDateUst is set in the "low level" layer, fEndDateUst is from previous cycle
     if (fEngine->Process(fBeginDateUst, fEndDateUst)) {
 
-        if (ResumeRefNum() < 0) {
-            jack_error("JackAudioDriver::ProcessGraphSyncMaster: ResumeRefNum error");
-        }
-
         if (ProcessReadSlaves() < 0) {
             jack_error("JackAudioDriver::ProcessGraphSync: ProcessReadSlaves error, engine may now behave abnormally!!");
         }
 
-        if (ProcessWriteSlaves() < 0) {
-            jack_error("JackAudioDriver::ProcessGraphSync: ProcessWriteSlaves error, engine may now behave abnormally!!");
+        if (ResumeRefNum() < 0) {
+            jack_error("JackAudioDriver::ProcessGraphSyncMaster: ResumeRefNum error");
         }
 
         // Waits for graph execution end
         if (SuspendRefNum() < 0) {
             jack_error("JackAudioDriver::ProcessGraphSync: SuspendRefNum error, engine may now behave abnormally!!");
+        }
+
+        if (ProcessWriteSlaves() < 0) {
+            jack_error("JackAudioDriver::ProcessGraphSync: ProcessWriteSlaves error, engine may now behave abnormally!!");
         }
 
     } else { // Graph not finished: do not activate it


amiartus-adit-jv:
Dear Jack2 Maintainers,

I have a question about JackAudioDriver.cpp. It seems to me like there could be a possible race condition in JackAudioDriver::ProcessGraphSyncMaster(), where the driver would do read and write on a slave device without waiting for graph execution end. What do you think?

best regards,

Adam

sletz:
What is the issue you're trying to fix in the first place ?

amiartus-adit-jv:
The end goal is to modify jack2 to allow use of alsa as a slave device and then modify async sequence to do following:
1.	Read master
2.	Read slave
3.	Write master
4.	Write slave
5.	Process graph

I would like to assure all reads are done before writes, as otherwise there could be additional latency introduced that reduces the possible time left over for processing clients.

sletz:
I still don't understand; your patch is modifying the `JackAudioDriver::ProcessGraphSyncMaster() `methods right ? Then why are you speaking about "modify async sequence" ?

amiartus-adit-jv:
right, this was discovered when I was doing the changes for async, there would be more patches modifying async sequence and alsa driver later.

I would like to know if my understanding of the code is correct, before I proceed further.

sletz:
I think changing this code is highly dangerous, since it will change the behavior of the whole graph scheduling.
And to go back on the initial problem: what is  issue with "using the alsa as a slave device" ? and when you say  "otherwise there could be additional latency introduced that reduces the possible time left over for processing clients" ?

amiartus-adit-jv:
the use case is to be able to use use multiple alsa devices with jack, we have considered multiple approaches:

1. using alsa multi plugin, however this is not available on all platforms we are using like qnx
2. alsa_in alsa_out, these introduce additional overhead due to additional threads and async sample rate conversion, as all our devices run on the same clock, this is not required for our use case
3. add additional parameters to the alsa driver that would allow use of multiple devices, at the end was not considered generic

therefore the chosen approach is to allow use of alsa driver as both master and slave device, this should allow more portability and configuring which devices are started

with regards to latency, the measurements I did have not revealed any issue, lets not focus on it further, sorry for confusion

sletz:
1) so why is the title "possible race question " ? Since I understand the issue is about using alsa driver in both master and slave modes?

2) "therefore the chosen approach is to allow use of alsa driver as both master and slave device, this should allow more portability and configuring which devices are started" : then why not code in the alsa driver itself about "master and slave" modes? And not modify the graph scheduling model (since as I said this seems highly suspicious as would need much more deep thinking to be sure the semantic is not modified).

amiartus-adit-jv:
I wanted to make sure there would not be any issues due to current processing model when we start adding multiple alsa devices.

When we add multiple alsa slave devices, I think it would make more sense to do all reads from alsa devices first (both master and slave), and then all writes. However, maybe there is a reason why it is done differently in current model, (read/write master first then slaves).

Therefore we would need to read from slave devices before processing of clients is started, as done in this patch. This patch is mainly to ask a question whether reading from slaves before processing clients is valid approach, and why the current model is doing it differently.

sletz:
Then I think the better and safer way is 1) do not change the current model 2) make the alsa driver support multiple-devices internally

amiartus-adit-jv:
when comparing to jack1 the sequence is following in sync mode:

1. read the slave drivers
2. read master driver
3. engine processing
4. write to slave drivers
5. write to master driver

this is different in jack2 sync mode, is there any reason why this was changed?

falkTX:
One thing to note is that soon I will be porting the zita-a2j/j2a internal client of jack1 into jack2, so you will be able to use multiple devices this way too.

amiartus-adit-jv:
thanks @falkTX, I could not find detailed information about this client, what would be the difference to https://github.com/jackaudio/jack2/blob/develop/example-clients/alsa_in.c ?

falkTX:
alsa_in/out are external applications and clients, while the zita stuff got merged into jack directly.
its code lives at https://github.com/jackaudio/tools/tree/master/zalsa

amiartus-adit-jv:
thank you for the information. at the end we decided to go for a different approach and are developing patches to jack2 alsa_driver.

therefore this PR is no longer valid.

PR #495, opened 2019-09-01
1.9.14 on MinGW/waf + build guide

lucianoiam:
Here is a patch that allows building latest JACK on MinGW using the waf scripts. No source code changes are involved except for a minor issue in JackMetadata.cpp

Also included:

- Detailed guide for setting up the dev environment.
- Updated Visual Studio project for building JackRouter (64bits only)

The patches are the less invasive as possible. All comments and testing welcomed.


windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+
+Finally for portaudio the Steinberg ASIO SDK should be manually downloaded.
+It can be found at https://www.steinberg.net/en/company/developers.html. The waf
+script will later check if the SDK is present at /opt/asiosdk
+
+wget https://www.steinberg.net/asiosdk -O /tmp/asiosdk.zip
+unzip /tmp/asiosdk.zip -d /tmp
+mv /tmp/asiosdk_{version}_{date} /opt/asiosdk
+
+The PKGBUILD description file needs to be patched so configure is called with
```
Schroedingers-Cat: I don't see any file like that in the /opt/asiosdk directory or in the jack repo. What file are you referring to?

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+the default version shipped by MinGW, it seems better idea to build from source
+to avoid any linker and runtime issues. A good technical explanation and/or
+evidence for this statement is currently missing from this guide.
+
+Fortunately there are PKGBUILD files for doing so together with a nice guide at
+https://github.com/msys2/MINGW-packages
+
+git clone https://github.com/msys2/MINGW-packages.git
+
+cd MINGW-packages/mingw-w64-{libname}-git
```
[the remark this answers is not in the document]
iamluciano: You are right, -git only applies to libtre, I agree on removing -git from the guide.

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+Some libraries like libsndfile and libsamplerate will ask for installing extra
+dependencies, it is ok to do so. The following need to be built in order:
+
+libtre
+libsystre
+
+libsystre is a wrapper around libtre that allows including <regex.h> later.
+Before building it adjust the includes path, it can be unset afterwards:
+
+export C_INCLUDE_PATH=/mingw64/include
```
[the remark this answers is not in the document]
iamluciano: I agree, it helps when doing things step by step.

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+
+db
+libsndfile
+libsamplerate
+
+Some libraries like libsndfile and libsamplerate will ask for installing extra
+dependencies, it is ok to do so. The following need to be built in order:
+
+libtre
+libsystre
```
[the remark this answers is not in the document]
iamluciano: Also agree it will look better

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+
+export C_INCLUDE_PATH=/mingw64/include
+
+Finally for portaudio the Steinberg ASIO SDK should be manually downloaded.
+It can be found at https://www.steinberg.net/en/company/developers.html. The waf
+script will later check if the SDK is present at /opt/asiosdk
+
+wget https://www.steinberg.net/asiosdk -O /tmp/asiosdk.zip
+unzip /tmp/asiosdk.zip -d /tmp
+mv /tmp/asiosdk_{version}_{date} /opt/asiosdk
```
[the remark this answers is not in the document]
iamluciano: I cannot recall if `/opt `is already present in a fresh MinGW install. `mkdir /opt/asiosdk -p `would help in if `/opt` is not already present but then the `mv` line will result in a possibly unnecessary level:

`/opt/asiosdk/asiosdk_{version}_{date}`

Also I am removing `{version}_{date}` on purpose so it is not carried to subsequent steps and also because the wscript checks at a hardcoded place` /opt/asiosdk`. Otherwise the user would have to adjust wscript depending the ASIO SDK minor version or date, or use a environment variable, etc.  I think this can be easily avoided by stripping the version/date suffixes right on this early filesystem step.

windows/README_MINGW
iamluciano: The paragraph talks about portaudio (Finally for portaudio...), so it is portaudio's PKGBUILD from the MINGW-packages. I will make that more clear.

windows/README_MINGW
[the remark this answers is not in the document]
Schroedingers-Cat: `/opt` is not present in a fresh MinGW install. Because of the version path situation you explained, it might be sufficient to change the line to `mkdir /opt`.

windows/README_MINGW
Schroedingers-Cat: Alright, that makes sense. Would be great to directly name this dependency like the others. 

Maybe it makes the instructions clearer if the modifications for the portaudio's PKGBUILD are placed before building any of the packages and then simply adding the `portaudio` library name to the [merged](https://github.com/jackaudio/jack2/pull/495/files/397bf247522c7d80c65d9486faf637132e2a1076?file-filters%5B%5D=No+extension#r322221335) list.

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+
+The PKGBUILD description file needs to be patched so configure is called with
+the necessary flags for ASIO:
+
+--with-asiodir=/opt/asiosdk   <- new option     
+--with-winapi=wmme,directx,wasapi,vdmks,asio   <- append 'asio' to existing
+
+Currently the shared library version of portaudio fails to build, a further
+patch is needed so only the static version is generated. Comment out lines
+[61-75], [82-83] from PKGBUILD.
```
Schroedingers-Cat: If you add
```
find /opt/asiosdk -name "*.lo" -type f -delete
find /opt/asiosdk -name "*.o" -type f -delete
```

between line 60 and 61 so that the build block looks like this
```
build() {
  export lt_cv_deplibs_check_method='pass_all'

  [[ -d "build-static-${MINGW_CHOST}" ]] && rm -rf "build-static-${MINGW_CHOST}"
  mkdir -p "${srcdir}/build-static-${MINGW_CHOST}"
  cd "${srcdir}/build-static-${MINGW_CHOST}"

  ../${_realname}/configure \
    --prefix=${MINGW_PREFIX} \
    --build=${MINGW_CHOST} \
    --host=${MINGW_CHOST} \
    --target=${MINGW_CHOST} \
    --disable-shared \
    --enable-static \
    --with-dxdir=${MINGW_PREFIX}/${MINGW_CHOST} \
    --with-asiodir=/opt/asiosdk \
    --with-winapi=wmme,directx,wasapi,wdmks,asio

  make

  find /opt/asiosdk -name "*.lo" -type f -delete
  find /opt/asiosdk -name "*.o" -type f -delete

  [[ -d "build-shared-${MINGW_CHOST}" ]] && rm -rf "build-shared-${MINGW_CHOST}"
  mkdir -p "${srcdir}/build-shared-${MINGW_CHOST}"
  cd "${srcdir}/build-shared-${MINGW_CHOST}"

  ../${_realname}/configure \
    --prefix=${MINGW_PREFIX} \
    --build=${MINGW_CHOST} \
    --host=${MINGW_CHOST} \
    --target=${MINGW_CHOST} \
    --enable-shared \
    --with-dxdir=${MINGW_PREFIX}/${MINGW_CHOST} \
    --with-asiodir=/opt/asiosdk \
    --with-winapi=wmme,directx,wasapi,wdmks,asio

  make
}
```
building both the shared and the static version works over here.

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+--------------
+
+- Clone repo
+
+git clone https://github.com/jackaudio/jack2
+
+- Build and install
+
+cd jack2
+./waf configure --prefix=/opt/jack && ./waf -p install
```
Schroedingers-Cat: The first `./waf` command fails here. From the config.log:
```
err: ../test.cpp:1:10: fatal error: opus/opus_custom.h: No such file or directory
 #include <opus/opus_custom.h>
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
This happens nine times for the libraries/headers `opus`, `readline`, `systemd`, `db`, `alloca`, `poll`, `execinfo` and `regex`. I tried to build and install them from source as the ones above - is that the right direction?

Building `readline` from source fails with:
```
../../readline-8.0/colors.c: In function '_rl_print_color_indicator':
../../readline-8.0/colors.c:178:27: warning: implicit declaration of function 'S_ISLNK'; did you mean 'S_ISBLK'? [-Wimplicit-function-declaration]
   else if (linkok == 0 && S_ISLNK(mode) && _rl_color_indicator[C_ORPHAN].string != NULL)
                           ^~~~~~~
                           S_ISBLK
```
How did you fix this?

windows/README_MINGW
iamluciano: Agree

windows/README_MINGW
iamluciano: Awesome finding, thanks

windows/README_MINGW
iamluciano: Never happened to me. Let me build from scratch following the guide and see. Thanks.

iamluciano: It works here, however I found three issues in the guide. Please let me know if the following helps:

1. If `pacman -Suy` ends with the following warning, you need to close the terminal AND run `pacman -Suy` again:

```
warning: terminate MSYS2 without returning to shell and check for updates again
warning: for example close your terminal window instead of calling exit
```

So please check there are no pending updates in your installation by running `pacman -Suy` again

2. This should be set before building any library:

`export C_INCLUDE_PATH=/mingw64/include`

3. Note the guide calls for cloning the official repository, which does not contain the wscript patches in this PR. Instead of cloning https://github.com/jackaudio/jack2 use fork  https://github.com/iamluciano/jack2, branch `target-mingw`.

I confirm your portaudio PKGBUILD patch works and allows building static and dynamic libraries.

Will update the guide soon, thanks for testing and sharing feedback.

iamluciano: Also note no dependencies are needed beyond those currently indicated in the guide. Missing headers is a symptom of something else.

iamluciano: There is a possibility the compiler toolchain installation fails in a not very explicit way, so I added details and an alternate solution for that step:

(Line 58 onwards)

That will decompress to a folder called `mingw64` in the working directory.
Now replace the files from the previous` mingw-w64-x86_64-toolchain` package
installation:

```
mv /mingw64 /mingw64.bak   -> backup original
rm /mingw64                -> see explanation below
mv mingw64 /               -> sjlj toolchain
```

The first step results in a file called `mingw64` file being created in `/`. The
file is not visible to the Windows File Explorer when inspecting `c:\msys64`, so
it is probably an artifact and safe to delete it. Otherwise the last `mv` step
will fail with:

`mv: cannot overwrite non-directory '/mingw64' with directory 'mingw64'`

An alternate solution consists in closing the MSYS2 window and replicating the
above procedure using the File Explorer. The following assumes the toolchain 7z
file was decompressed to the home directory:

Rename `c:\msys64\mingw64` to `c:\msys64\mingw64.bak`
Move `c:\msys64\home\{username}\mingw64` to `c:\msys64`

Schroedingers-Cat: >     3. Note the guide calls for cloning the official repository, which does not contain the wscript patches in this PR. Instead of cloning https://github.com/jackaudio/jack2 use fork  https://github.com/iamluciano/jack2, branch `target-mingw`.
That was the problem. Could have thought of that. So now the first command works.  

However, calling `./waf -p install` fails and I'm not sure where to look for an error message. I tried running this in verbose and it seems that the error is coming from `collect2.exe` (line 27 in the pastebin link), but I'm not sure if there is any more info. Here is a pastebin of the terminal output: https://pastebin.com/nBFEb89r  
Any idea where to dig for more info or what is causing this?

iamluciano: Looks like the portaudio library is not included during link. Can you double check portaudio is building fine and the resulting .xz package is properly installed?

```
cd ~/MINGW-packages/mingw-w64-portaudio
pacman -U *.xz
```

Then check if waf configure is seeing portaudio:

`Checking for 'portaudio-2.0 >= 19'       : yes`

Schroedingers-Cat: Strange, I'm sure I already did that. Re-doing these steps fixed it. Nevermind!

Schroedingers-Cat: Strange, I'm sure I already did that. Re-doing these steps fixed it. Nevermind!

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+windows/JackRouter/JackRouter_MinGW_deps.vcxproj
+
+Differences from the original JackRouter.vcxproj:
+
+- Depends on the asiosdk files from the previous MinGW JACK installation (i.e.,
+  it points to absolute paths starting with c:\msys64\opt)
+- Links to the jack.dll.a created by the previous build, instead of libjack.lib
+  This means JackRouter.dll will currently depend on libjack-0.dll
+- Force includes stdint.h and defines _STDIN_H to avoid int8_t, int32_t and
+  uint32_t basic types redefinition duringn compilation
```
Schroedingers-Cat: Using VS2017 15.9.16 and the Windows SDK 10.0.17134.0, I'm unable to build the project found in `windows\JackRouter\JackRouter_MinGW_deps.vcxproj` (from your forked repository) due to redefinition of int8_t, int32_t and uint32_t. Configuration is `Debug` and `x64`. Interestingly, I'm unable to find the mentioned `_STDIN_H` in the entire project solution with the VS search function.

iamluciano: Sorry, there is an unfortunate typo in the guide, it is `_STDINT_H` and not `_STDIN_H`:

```
$ grep -R "_STDINT_H" .
./JackRouter_MinGW_deps.vcxproj:      <PreprocessorDefinitions>_STDINT_H;WIN32;NDEBUG;_WINDOWS;PSAPI_VERSION=2;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```

By the time I wrote the guide I used an existing VS installation. I still have not tried to build JackRouter totally from scratch (including installing VS), and this `_STDINT_H` thing is really a quick hack rather than a final solution. The latter would involve building JackRouter using a waf script, but that might not be trivial because it involves transitioning from MSVC to GCC.

Let me retry as soon as I get to a Windows machine again. Meanwhile you can try commenting out any redefinition of `int8_t`, `int32_t` and `uint32_t`. That is ugly as sin but should work.

Schroedingers-Cat: The preprocessor definiton was missing in the `Debug` configuration (which is the default VS setting when opening a project for the first time). The configuration `Release` works. I fixed these settings with this commit https://github.com/Schroedingers-Cat/jack2/commit/91d54ecca3fa603d2742de53bc96fd77eff6c282 but I'm unable to PR into your fork. You might also want to consider including this commit https://github.com/Schroedingers-Cat/jack2/commit/91d54ecca3fa603d2742de53bc96fd77eff6c282 which ignores VS generated user files and build files from version control.  

Additionally, you'll need to install the two marked entries from the VS installer in order to succesfully build: 
grafik

The left circled selection is rather obvious but the one to the right is an additional selection you have to make manually.

windows/README_MINGW
```diff
@@ -0,0 +1,218 @@
...
+
+Tested working clients:
+Ardour
+Bitwig Studio
+QJackCtl
+JackRouter
+
+Example of starting the JACK server including MIDI support for a Focusrite USB
+audio device using ASIO:
+jackd -R -S -X winmme -d portaudio -p 32 -r 48000 -d "ASIO::Focusrite USB ASIO"
```
Schroedingers-Cat: My jack setup, which is working fine with 1.9.11, crashes upon starting: 
```
08:47:44.103 Statistics reset.
08:47:55.518 JACK is starting...
08:47:55.519 ./jackd.exe -S -dportaudio -r48000 -p1024 -D -C"ASIO::ASIO HDSPe FX" -P"ASIO::ASIO HDSPe FX"
08:47:55.573 JACK was started with PID=78262880.
08:47:55.586 JACK was stopped
08:47:55.587 JACK has crashed.
08:47:57.723 Could not connect to JACK server as client. - Overall operation failed. Please check the messages window for more info.
```
The issue might be that the instructions from line 168-178 didn't exactly tell me where to put the files that were not mentioned in lines 184-189. So where to put libsndfile-1, libsamplerate-0 and libdb-6.0?

iamluciano: You can place the DLLs listed in 168-178 in the same directory the binaries will finally reside in your disk, for example `C:\Program Files\Jack`. Windows default behavior is to first look for DLLs in the same directory the invoked EXEs are located, then in system-wide directories like `C:\Windows`, `C:\Windows\System32`, and a couple more. The guide does not dig too much into deployment details because this should be either addressed by a waf script update, or better a wizard-like installer which will be more familiar to Windows users, like the current binary distribution implements. There is a recommendation to place the libraries listed in 184-189 in `C:\Windows` because those are dependencies of `libjack-0.dll` _only_, and any Jack client will indirectly need to link against them (in contrast the first DLL list is for `jackd.exe` et al.) . It could be any other system-wide DLLs directory, I just followed what the current binary distribution does. Do not forget to rename `libjack-0.dll` to `libjack64.dll` inside `C:\Windows` otherwise clients will not be able to find it.

I think the guide can be improved by adding an example of how to install everything to Program Files, until an automated solution is developed.

If it continues to crash then something else is happening. Btw, where did you take the log excerpt from?

windows/README_MINGW
iamluciano: Awesome finding － yes, I forgot I only worked on the Release variant. Thanks for the update, maybe you cannot initiate the PR because you forked the original Jack2 repo and not my own fork? . It is correct MFC needs to be installed in order for the build to succeed, I am not sure why (I am not a proficient Windows programmer) but think I read that somewhere and did not include the step in the guide, so will update accordingly. Thanks.

Schroedingers-Cat: Turns out that my fork was a bit outdated and that's propably why github hid your repo when creating a PR. That's now fixed and you can just accept the PR and have the changes included here automatically.

iamluciano: Merged, thanks.

windows/README_MINGW
Schroedingers-Cat: > 
> 
> You can place the DLLs listed in 168-178 in the same directory the binaries will finally reside in your disk, for example `C:\Program Files\Jack`. 

Cool, that works. Awesome! I can also confirm that this works with the latest VS2019 and toolchain+SDK.
I know about the default behaviour, it just wasn't clear from the description in I had less time for trying things out yesterday. You might want to include that info.

I'm surprised that the libraries from 171-178 are different from line 184-189. Any reason why? I'm working on a jack client for unity. With the binaries generated from this guide, that jack client only works if I copy over all of the dlls from line 171-178 to `c:\windows` while jack-enabled applications should just need the dlls from line 184-189. My jack client works fine using the 1.9.11 binary release.

iamluciano: Great to know it also works with VS2019. I updated the guide to reflect that and clarify the paths selection. I am not really familiar with the JACK internals, but think libjack only needs a subset of the server dependencies, for example it does not need to interface to the audio device hence portaudio is not needed by libjack. That explains the discrepancies between DLL lists. I am also not a proficient Windows programmer, but know that cluttering the Windows directory is not a friendly thing to do. I think libjack should be either statically linked to its dependencies or moved to its own place and made available to clients somehow (if that is possible at all). Then the server + standalone binaries + dependency DLLs are not a problem if placed in their own directory. The updated guide shows how to do that using` C:\Program Files\Jack` as an example.

Thanks for all the contributions, I am pushing the updated guide now.

windows/README_MINGW
```diff
@@ -0,0 +1,268 @@
...
+An alternate solution consists in closing the MSYS2 window and replicating the
+above procedure using the File Explorer. The following assumes the toolchain 7z
+file was decompressed to the home directory:
+
+Rename C:\msys64\mingw64 to C:\msys64\mingw64.bak
+Move C:\msys64\home\{username}\mingw64 to C:\msys64
+
+Make sure gcc runs and it is the expected version:
+
+$ gcc --version
```
auphofBSF: I have to put  a path prefix ` /mingw64/bin/` ie ` /mingw64/bin/gcc --version`  should there be an env Path update ?

windows/README_MINGW
```diff
@@ -0,0 +1,268 @@
...
+--------------
+
+- Clone repo
+
+git clone https://github.com/jackaudio/jack2
+
+- Build and install
+
+cd jack2
+./waf configure --prefix=/opt/jack && ./waf -p install
```
auphofBSF: Note to possibly add for a mingw build.  
The `./waf configure .....`  was not finding the new built libraries.  I had an msys64 pkg-config , `which pkg-config`   showed `/usr/bin/pkg-config`   if i ran ` /mingw64/bin/pkg-config --list-all` it showed all the built libraries. 
My fix  was to ensure mingw pkg-config was installed `pacman -S mingw-w64-x86_64--pkg-config`   and to remove msys pkg-config `pacman -R pkg-config`

windows/README_MINGW
```diff
@@ -0,0 +1,216 @@
...
+
+if bld.env['IS_WINDOWS']:
+    clientlib.env.append_value('LINKFLAGS', ['-static-libstdc++', '--disable-auto-import'])
+
+- Finally build and install
+
+./waf configure --prefix=/opt/jack --db=no
+./waf
+./waf install
+
```
auphofBSF: Unfortunately can not get past this, any suggestions please, have setup asio as you have defined

```
$ ./waf configure --prefix=/opt/jack --db=no
Setting top to                           : /home/BSFau/jack2
Setting out to                           : /home/BSFau/jack2/build
Checking for 'clang++' (C++ compiler)    : not found
Checking for 'g++' (C++ compiler)        : /mingw64/bin/g++
Checking for 'clang' (C compiler)        : not found
Checking for 'gcc' (C compiler)          : /mingw64/bin/gcc
Checking platform                        : Windows
Checking for ASIO SDK                    : yes
Checking for program 'pkg-config'        : /mingw64/bin/pkg-config
Checking for 'alsa >= 1.0.18'            : not found
Checking for 'libffado >= 1.999.17'      : not found
Checking for 'gtkIOStream >= 1.4.0'      : not found
Checking for 'eigen3 >= 3.1.2'           : not found
Checking for header windows.h            : yes
Checking for 'portaudio-2.0 >= 19'       : yes
Checking for header mmsystem.h           : yes
Checking for 'celt >= 0.11.0'            : not found
Checking for 'celt >= 0.8.0'             : not found
Checking for 'celt >= 0.7.0'             : not found
Checking for 'celt >= 0.5.0'             : not found
Checking for header opus/opus_custom.h   : not found
Checking for 'opus >= 0.9.0'             : not found
Checking for 'samplerate'                : yes
Checking for 'sndfile'                   : yes
Checking for library readline            : not found
Checking for header readline/readline.h  : not found
Checking for header systemd/sd-daemon.h  : not found
Checking for library systemd             : not found
Checking for header alloca.h             : not found
Checking for ppoll                       : no
Checking for header execinfo.h           : not found
Invalid argument 'function_name' in test
Checking for header regex.h              : yes
Invalid argument 'function_name' in test
Checking for header winsock2.h           : yes
Invalid argument 'function_name' in test
Checking for header ['windows.h', 'mmsystem.h'] : yes
Invalid argument 'function_name' in test
Checking for header ['windows.h', 'psapi.h']    : yes
Checking for ucontext->uc_mcontext.gp_regs      : no
Checking for ucontext->uc_mcontext.uc_regs      : no
Checking for ucontext->uc_mcontext.mc_gregs     : no
Checking for ucontext->uc_mcontext.gregs        : no
Checking for NGREG                              : no

==================
JACK 1.9.13 svn revision will checked and eventually updated during build
Maximum JACK clients                            : 64
Maximum ports per application                   : 768
Install prefix                                  : /opt/jack
Library directory                               : /opt/jack/lib
Drivers directory                               : /opt/jack/bin/jack
Build debuggable binaries                       : no
C compiler flags                                : ['-Wall']
C++ compiler flags                              : ['-Wall', '-std=gnu++11']
Linker flags                                    : ['-Wl,--enable-auto-import', '-Wl,--enable-auto-import']
Build with engine profiling                     : no
Build with 32/64 bits mixed mode                : no
Build standard JACK (jackd)                     : yes
Build D-Bus JACK (jackdbus)                     : no
Autostart method                                : classic
Build doxygen documentation                     : no
Enable ALSA driver                              : no
Enable FireWire driver (FFADO)                  : no
Enable IIO driver                               : no
Enable Portaudio driver                         : yes
Enable WinMME driver                            : yes
Build with CELT                                 : no
Build Opus netjack2                             : no
Build with libsamplerate                        : yes
Build with libsndfile                           : yes
Build with readline                             : no
Use systemd notify                              : no
Use Berkeley DB (metadata)                      : no

'configure' finished successfully (14.038s)

BSFau@BSF-PC-AU1 MSYS ~/jack2
$ ./waf -v
Waf: Entering directory `/home/BSFau/jack2/build'
[138/223] Linking build/common/audioadapter.dll
00:24:09 runner ['/mingw64/bin/g++', '-Wl,--enable-auto-import', '-Wl,--enable-auto-import', '-Wl,--out-implib,common/audioadapter.dll.a', '-shared', 'common/JackResampler.cpp.7.o', 'common/JackLibSampleRateResampler.cpp.7.o', 'common/JackAudioAdapter.cpp.7.o', 'common/JackAudioAdapterInterface.cpp.7.o', 'common/JackAudioAdapterFactory.cpp.7.o', 'windows/portaudio/JackPortAudioAdapter.cpp.7.o', 'windows/portaudio/JackPortAudioDevices.cpp.7.o', '-o/home/BSFau/jack2/build/common/audioadapter.dll', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-Lcommon', '-LD:/msys64/mingw64/lib', '-LD:/msys64/mingw64/lib', '-ljackserver', '-lpthread', '-lregex', '-lws2_32', '-lpsapi', '-lwinmm', '-lsamplerate', '-lportaudio', '-lwinmm', '-lm', '-ldsound', '-lole32', '-lwinmm', '-lm', '-luuid', '-lsetupapi', '-lole32', '-lwinmm', '-lm', '-lole32', '-luuid', '-lwinmm', '-lm', '-lole32', '-luuid']
[155/223] Linking build/jack_portaudio.dll
00:24:09 runner ['/mingw64/bin/g++', '-Wl,--enable-auto-import', '-Wl,--enable-auto-import', '-Wl,--out-implib,portaudio.dll.a', '-shared', '-shared', 'windows/portaudio/JackPortAudioDevices.cpp.8.o', 'windows/portaudio/JackPortAudioDriver.cpp.8.o', '-o/home/BSFau/jack2/build/jack_portaudio.dll', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-Lcommon', '-LD:/msys64/mingw64/lib', '-ljackserver', '-lportaudio', '-lwinmm', '-lm', '-ldsound', '-lole32', '-lwinmm', '-lm', '-luuid', '-lsetupapi', '-lole32', '-lwinmm', '-lm', '-lole32', '-luuid', '-lwinmm', '-lm', '-lole32', '-luuid', '-lpthread', '-lregex', '-lws2_32', '-lpsapi', '-lwinmm']
windows/portaudio/JackPortAudioDevices.cpp.8.o:JackPortAudioDevices.cpp:(.text+0x1821): undefined reference to `PaAsio_GetAvailableBufferSizes'
windows/portaudio/JackPortAudioDevices.cpp.8.o:JackPortAudioDevices.cpp:(.text+0x1bfd): undefined reference to `PaAsio_GetAvailableBufferSizes'
windows/portaudio/JackPortAudioDriver.cpp.8.o:JackPortAudioDriver.cpp:(.text+0x11e3): undefined reference to `PaAsio_GetInputChannelName'
windows/portaudio/JackPortAudioDriver.cpp.8.o:JackPortAudioDriver.cpp:(.text+0x1316): undefined reference to `PaAsio_GetOutputChannelName'
collect2.exe: error: ld returned 1 exit status

windows/portaudio/JackPortAudioDevices.cpp.7.o:JackPortAudioDevices.cpp:(.text+0x1821): undefined reference to `PaAsio_GetAvailableBufferSizes'
windows/portaudio/JackPortAudioDevices.cpp.7.o:JackPortAudioDevices.cpp:(.text+0x1bfd): undefined reference to `PaAsio_GetAvailableBufferSizes'
collect2.exe: error: ld returned 1 exit status

Waf: Leaving directory `/home/BSFau/jack2/build'
Build failed
 -> task in 'portaudio' failed with exit status 1:
        {task 7696569732816: cxxshlib JackPortAudioDevices.cpp.8.o,JackPortAudioDriver.cpp.8.o -> jack_portaudio.dll,portaudio.dll.a}
['/mingw64/bin/g++', '-Wl,--enable-auto-import', '-Wl,--enable-auto-import', '-Wl,--out-implib,portaudio.dll.a', '-shared', '-shared', 'windows/portaudio/JackPortAudioDevices.cpp.8.o', 'windows/portaudio/JackPortAudioDriver.cpp.8.o', '-o/home/BSFau/jack2/build/jack_portaudio.dll', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-Lcommon', '-LD:/msys64/mingw64/lib', '-ljackserver', '-lportaudio', '-lwinmm', '-lm', '-ldsound', '-lole32', '-lwinmm', '-lm', '-luuid', '-lsetupapi', '-lole32', '-lwinmm', '-lm', '-lole32', '-luuid', '-lwinmm', '-lm', '-lole32', '-luuid', '-lpthread', '-lregex', '-lws2_32', '-lpsapi', '-lwinmm']
 -> task in 'audioadapter' failed with exit status 1:
        {task 7696569673680: cxxshlib JackResampler.cpp.7.o,JackLibSampleRateResampler.cpp.7.o,JackAudioAdapter.cpp.7.o,JackAudioAdapterInterface.cpp.7.o,JackAudioAdapterFactory.cpp.7.o,JackPortAudioAdapter.cpp.7.o,JackPortAudioDevices.cpp.7.o -> audioadapter.dll,audioadapter.dll.a}
['/mingw64/bin/g++', '-Wl,--enable-auto-import', '-Wl,--enable-auto-import', '-Wl,--out-implib,common/audioadapter.dll.a', '-shared', 'common/JackResampler.cpp.7.o', 'common/JackLibSampleRateResampler.cpp.7.o', 'common/JackAudioAdapter.cpp.7.o', 'common/JackAudioAdapterInterface.cpp.7.o', 'common/JackAudioAdapterFactory.cpp.7.o', 'windows/portaudio/JackPortAudioAdapter.cpp.7.o', 'windows/portaudio/JackPortAudioDevices.cpp.7.o', '-o/home/BSFau/jack2/build/common/audioadapter.dll', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-Lcommon', '-LD:/msys64/mingw64/lib', '-LD:/msys64/mingw64/lib', '-ljackserver', '-lpthread', '-lregex', '-lws2_32', '-lpsapi', '-lwinmm', '-lsamplerate', '-lportaudio', '-lwinmm', '-lm', '-ldsound', '-lole32', '-lwinmm', '-lm', '-luuid', '-lsetupapi', '-lole32', '-lwinmm', '-lm', '-lole32', '-luuid', '-lwinmm', '-lm', '-lole32', '-luuid']

```

windows/README_MINGW
lucianoiam: I ran the guide twice and do not recall having to do that. I am not really experienced with MinGW and its build tools, all suggestions are welcomed (continues on next answer)

windows/README_MINGW
lucianoiam: Can you submit a patch for this and the previous (possible) issue?

windows/README_MINGW
lucianoiam: Is it possible your portaudio is missing ASIO support? the binary version that can be installed via pacman lacks ASIO. The guide contains instructions to build portaudio from the MinGW source package, its PKGBUILD file has to be patched to enable ASIO.

lucianoiam:
@Schroedingers-Cat  may I ask if you finally have been able to build JACK using this guide? I bet there are not many Windows/MinGW JACK users out there. Myself have not rebuilt JACK after the 1.9.14 update, in fact I have not booted Windows ever since.

windows/README_MINGW
auphofBSF: I believe I have patched  this successfully
image

I checked the build logs and to my relatively inexperienced eye it looked ok, but on checking again now i see in configuration summary `ASIO = no`' I will look further but if I have missed something please let me know. Thanks for your huge effort with this

image

auphofBSF: FIXED, FOUND it, my error I only added ASIO support to the build static,

can we use git apply with a patch instead of individually hand editing.  maybe something that can be scripted, but forgive me if this is not valid I am very much a  NOOB with mingW and C++,C let alone cross platform.

@lucianoiam  this PR is an awesome attempt  at simplifying the build, well done

by the way for anyone else hitting this a git diff should look like this
```
diff --git a/mingw-w64-portaudio/PKGBUILD b/mingw-w64-portaudio/PKGBUILD
index 4daa221b7..7ee63ee0c 100644
--- a/mingw-w64-portaudio/PKGBUILD
+++ b/mingw-w64-portaudio/PKGBUILD
@@ -53,10 +53,12 @@ build() {
     --disable-shared \
     --enable-static \
     --with-dxdir=${MINGW_PREFIX}/${MINGW_CHOST} \
-    --with-winapi=wmme,directx,wasapi,wdmks
+         --with-asiodir=/opt/asiosdk  \
+    --with-winapi=wmme,directx,wasapi,wdmks,asio

   make
-
+  find /opt/asiosdk -name "*.lo" -type f -delete
+  find /opt/asiosdk -name "*.o" -type f -delete
   [[ -d "build-shared-${MINGW_CHOST}" ]] && rm -rf "build-shared-${MINGW_CHOST}"
   mkdir -p "${srcdir}/build-shared-${MINGW_CHOST}"
   cd "${srcdir}/build-shared-${MINGW_CHOST}"
@@ -68,7 +70,8 @@ build() {
     --target=${MINGW_CHOST} \
     --enable-shared \
     --with-dxdir=${MINGW_PREFIX}/${MINGW_CHOST} \
-    --with-winapi=wmme,directx,wasapi,wdmks
+         --with-asiodir=/opt/asiosdk  \
+    --with-winapi=wmme,directx,wasapi,wdmks,asio

   make
 }

```

lucianoiam: That is great! this patch makes clear what has to be done in PKGBUILD. I think a clean future proof solution is submitting a patch to the MinGW package mantainers that turns on ASIO support by default for portaudio if the required dependencies are present, without hardcoding a path or properly telling the user where to place ASIO.

Schroedingers-Cat:
@lucianoiam yes, its working fine our all our machines :)

windows/README_MINGW
Schroedingers-Cat: I'm not entirely sure about this but I think it's safer to run
```bash
find /opt/asiosdk -name "*.lo" -type f -delete
find /opt/asiosdk -name "*.o" -type f -delete
```
also after the last `make`, as the temp files from the last build could cause issues when re-running this with the first build. What do you think?

lucianoiam: You are right, I am updating the guide now. Stay tuned for the next commit, it also contains a patch to allow compiling 1.9.14 on MinGW. Thanks.

lucianoiam:
Thanks for the feedback. Maybe the project maintainer @falkTX is interested in merging this? I submitted an update that allows targeting 1.9.14 on MinGW, there had been some updates in JackMetadata.cpp since this guide and necessary (minor) patches were originally written.

Basically it adds some `#ifdefs` **only for WIN32** that:

- Redefines `jack_server_dir` so it points to c:\windows\temp, thus providing valid absolute paths for the DB files. The redefinition is only done inside JackMetadata.cpp, not JackPlatformPlug_os.h because I do not know the implications of doing the latter. Maybe it is not the cleanest solution but it is unobtrusive.

- Redefines `mkdir()` as a macro because Win32 `mkdir()` only takes one parameter and the source code passes two. Therefore the mode is discarded.

Unrelated, but the CI build appears to be systematically failing for this and other contributors' patches with similar errors.

falkTX:
You cannot redefine `jack_server_dir`, that would not work well.
We can define a new macro that uses custom location in win32 or `jack_server_dir` everywhere else.
But using `c:\windows\temp` as hardcoded location is not okay, it needs to use `%temp%` or similar

lucianoiam:
The `jack_server_dir` redefinition is scoped to JackMetadata.cpp and works without side effects, but I agree that was a lazy hack. I pushed a cleaner solution that adds a new member variable in JackMetadata to hold the database files directory. On Win32 it is initialized to a temp path returned by the system, and to `jack_server_dir` on all other platforms to keep existing behavior. Then the class implementation references this variable instead of the `jack_server_dir` #define.

Schroedingers-Cat:
What is preventing this from being merged?

falkTX:
I am going to have a take at building jack with mingw soon.
When I do that I can review the PR.

szszoke:
@falkTX Is there something that I as someone who also uses Jack and Ardour on Windows can do?

falkTX:
Right, it is about time to check on this.
To be real honest, I dislike the approach of having to follow a bunch of manual instructions just to get something to build.
The changes for win32 compatibility are very welcome, the README is also welcome if it allows you to build it, but I would take a different approach on that.

My current plan is to get cross-compilation going for mingw, a project I setup at https://github.com/DISTRHO/PawPaw
This project has a single entry-point script which will build all required dependencies.
It is very similar to what I have been using for my Carla plugin host, but now in a common project that can be useful for more than just Carla (the actual target is to build cross-platform LV2 plugins)

So once I have all the dependencies that JACK needs, we should be able to just run a script and get statically-built JACK2 binaries (still dlls, but in a way that libpthread, libgcc etc are not needed)

I would be very glad to receive feedback on this idea.
For now the project needs a linux machine for cross-compilation, but I am open to try to make it work under cygwin or WSL.

lucianoiam:
First of all PawPaw looks great and the way to go! looking forward to see it powering JACK; I'll have a deeper look for my own upcoming projects.

Now there are official plans for Windows I think this guide is deprecated but let me defend it a little as it could be of interest to someone out there still.

There are two separate aspects here

1. Knowledge to create a build environment
2. Knowledge to build, package, distribute a project

The guide is lengthy because it mostly covers #1 and JACK does not provide what Windows users/devs expect which I guess are Visual Studio projects. I think I started it because some Python binding was failing on Windows using the binary distribution of JACK, so decided to try compiling everything from source and failed miserably (cannot recall why now). Later the resulting environment served to develop some metadata stuff.

I made it verbose on purpose because not only the whole process is alien to Windows users but I myself fear forgetting the little details and there are too many steps involved. Hey we had to wait until 2020 for MS to ~~steal~~ get inspired to write a package manager! https://www.theverge.com/2020/5/28/21272964/microsoft-winget-windows-package-manager-appget-copied

Being said that, nobody likes following lengthy guides and this should not be treated as anything other than a stopgap measure :)

Until PawPaw itself runs on some of the multiple Windows options (VS, Cygwin, WSL2, MinGW...) this guide might still be useful for users wanting to build without setting up a separate Linux machine. In such case I published it to a blog:

https://medium.com/@lucianoiam/how-to-build-jack-on-mingw-ad8bcb2c22dd

I think this PR should be closed now.

falkTX:
I fully agree with those 2 points.
So, while the PawPaw project is not fully ready to build JACK2 binaries, I am quite okay with merging this actually. The README is useful for explaining some basic concepts, as you say.

Even while the CI is not fully working at the moment, I can see that this PR does not break Linux builds.
So I will actually merge this.
Once I have found my way to build JACK2 on Windows, I will rename that readme or something.

Thanks everyone for the changes made here

szszoke:
Just a heads-up if someone is getting undefined reference errors to `ASIO` stuff while trying to compile `portaudio`:

I also needed to update the `prepare` function in the `PKGBUILD` file for `portaudio`.

I added these 👇 after each `make` call. Not sure if the last one is required but without this the compilation failed with undefined references to `ASIO` stuff.
```
find /opt/asiosdk -name "*.lo" -type f -delete
find /opt/asiosdk -name "*.o" -type f -delete
```

szszoke:
Another heads-up: I didn't have the `PKG_CONFIG_PATH` environment variable set up correctly so when I was running `./waf configure` it didn't find `portaudio` and I didn't have support for it.

In my case the I had it set to this:
```
export PKG_CONFIG_PATH=/mingw64/lib/pkgconfig
```

szszoke:
I hit a roadblock. After successfully compiling Jack, copying the 2 Jack DLLs and their dependencies to my Windows folder, I can't make Ardour detect that Jack is running. When I am creating a new session or trying to change the audio engine, Jack is simply not in the `Audio system` dropdown list. Only portaudio.

I made sure to uninstall every prior version of Jack from my system.

@lucianoiam Maybe you have a clue what could be going on here? Did you face anything similar when you were testing Jack?

lucianoiam:
Cannot recall something like that. Just tried now and JACK is correctly detected by Ardour (official v6 binaries) on my Windows machine which has a custom JACK build installed. I am wondering if the fact I ran the official JACK installer at least once in the past on this machine can make a difference, but what you wrote implies we have a similar setup. I do not know how Ardour actually detects a JACK installation; I would start by checking if this is an Ardour specific issue. Have you tried with any other clients? for example JACK's own command line tools.

szszoke:
I did in fact had Jack installed via the official installer before.

I tried `jack_cpu_load` which seemed to give me a value roughly each second. I will try some other clients.

In the meantime, can you double check that nothing is missing from the instructions?

lucianoiam:
The guide is basically a hack until an official build solution is released, it is possible something is missing for it. But since other users reported success following I cannot do much other than suggesting you try to isolate the error cause. We do know JACK2 builds and runs on your machine because `jack_cpu_load` works as expected; that is a great starting point.

PR #498, opened 2019-09-08
MIDI port pretty names

iamluciano:
This patch makes JACK set the pretty name for each MIDI port on server startup. Currently the pretty name is equal to the MIDI device name as reported by the system MIDI implementation. The following JACK MIDI drivers have been modified:

alsarawmidi, coremidi, winmme

The pretty names are not updated if a previous value is found, ie. to avoid overwriting user defined values. A better implementation might set a new "device name" key property instead, but existing clients already rely on pretty names for presenting ports to the user (for example QjackCtl and Ardour).

There is an existing property key presumably for describing hardware (http://jackaudio.org/metadata/hardware). This patch also sets such property to the MIDI device name on server startup, overwriting any previous value found. If future clients rely on a hardware/device property on addition to pretty names for presentation, pretty names might be reserved for custom names exclusively (ie., user defined).

The ultimate goal is to add a practical solution for MIDI port names confusion on systems with a large quantity of such devices, which is not uncommon in home studio environments.

### DIFF
--- a/common/JackEngine.cpp
+++ b/common/JackEngine.cpp
@@ -1073,7 +1073,7 @@ int JackEngine::PortRename(int refnum, jack_port_id_t port, const char* name)
     return 0;
 }
 
-int JackEngine::PortSetPrettyNameProperty(jack_port_id_t port, const char* pretty_name)
+int JackEngine::PortSetDeviceName(jack_port_id_t port, const char* pretty_name)
 {
     jack_uuid_t uuid = jack_port_uuid_generate(port);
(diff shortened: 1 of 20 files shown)


falkTX:
Very cool, thanks a lot.
The only thing I would change is to give a different name to `int JackEngine::PortSetDeviceName` as it does not actually do that, it only sets metadata.
But I can do so after merging.

Audio ports deserve a similar treatment. there is a branch in jack2 that has code that exposes those, with meta-data it makes sense to put the names there.

PR #501, opened 2019-09-24
Extend API to support S16 and S32 audio formats

twischer-adit:
This new API allows applications to exchange the audio data with JACK as S16 or S32 format. It can easily extended for other formats.
The conversion is done by the already optimized functions implemented in `common/memops.c`

### DIFF
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -26,6 +26,9 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 #include "JackGlobals.h"
 #include "JackTime.h"
 #include "JackPortType.h"
+#include "jack.h"
+#include "memops.h"
+#include <typeinfo>
 #include <math.h>
 #define __STDC_FORMAT_MACROS 1
 #include <inttypes.h>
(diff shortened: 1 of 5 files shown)


twischer-adit:
@falkTX 
Thanks a lot for your fast feedback.

> but I dont really see the point of this.

We have multiple internal and external JACK clients which process the audio data in S16 (e.g. a wav/mp3 player). I would like to reuse the well optimized memops conversion functions but I do not want to duplicate this implementation for each JACK client.
I thought exporting the memops functions would not a good approach. With this solution the internal data type of JACK is still kept at float but on client site a conversion can easily be done.

A third option which I had in my mind was extending the jack_port_register() API function and provide `JACK_S16_AUDIO_TYPE` and `JACK_S32_AUDIO_TYPE` in addition to `JACK_DEFAULT_AUDIO_TYPE` for the `port_type` argument of this function.
But this would change the internal behaviour of JACK (e.g. conversion has done after mixing audio data). So I would expect this would highly increase the complexity of JACK.
The solution I chose only implements something on top of the JACK API.

Do you prefer any of the two other solutions (publish memops functions via API or extend jack_port_register())?
Or do you have another idea how I could provide a generic format conversion feature?
(I would expect there exists already a lot of software which currently has to reimplement the format conversion and has to optimize it by its own)

falkTX:
Honestly I do not think either of those is a good idea.
Using JACK for music players seems the wrong approach to begin with, pulseaudio is more suited for that.
maximum that I see could be done is expose the converter functions in their own API, like the ringbuffer does.

twischer-adit:
@falkTX 
> Using JACK for music players seems the wrong approach to begin with, pulseaudio is more suited for that.

Yes for sure when we would do only music playback. But in addition we also have JACK clients for equalizer, echo noise cancellation, ... (which have quite low latency requirements like a bluetooth phone call).

> maximum that I see could be done is expose the converter functions in their own API, like the ringbuffer does.

That is a great idea. I will move it to a new source and header file. Would you like to reopen this PR than or should I create a new PR?

falkTX:
you do not need to move any code, just export the needed functions and defined them in a new header.
it is better to open a new, clean PR.
but note that if this gets merged, is not going on next planned release, only after

twischer-adit:
@falkTX
> but note that if this gets merged, is not going on next planned release, only after

Yes. That is fine for me. I only want to align this upstream repository as far as possible with our internal repository.

Updated PR can be found in https://github.com/jackaudio/jack2/pull/503

PR #503, opened 2019-09-25
Provide additional API to support S16 and S32 audio formats

twischer-adit:
This new API allows applications to exchange the audio data with JACK as S16 or S32 format. It can easily extended for other formats.

### DIFF
--- a/common/JackFormatConverter.cpp
+++ b/common/JackFormatConverter.cpp
@@ -20,11 +20,16 @@
 */
 
 #include <cstring>
+#include <math.h>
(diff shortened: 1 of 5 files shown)


falkTX:
For the sake of not leaving PRs open forever, I have to say these changes are too drastic for the core aspects of jack.
So far jack has always assumed audio data is in 32bit floating-point type. It is hard for me to see the benefit of this change.
Also, the header introduced in this PR is not C compatible.

So I am closing this one as "wont fix".

PR #505, opened 2019-09-29
Set initial metadata for ALSA MIDI raw and seq ports

iamluciano:
This patch extends https://github.com/jackaudio/jack2/pull/498 by setting initial metadata also for ALSA `raw` and `seq` MIDI ports.

@falkTX what do you think about renaming` PortSetDeviceMetadata()` to `PortSetInitialMetadata()` ? I think the latter better explains what that method does. There is another lowercase version `jack_port_set_device_metadata()` in `alsa_midi_jackmp.cpp` that should also be renamed, just in case.

### DIFF
--- a/linux/alsa/JackAlsaDriver.cpp
+++ b/linux/alsa/JackAlsaDriver.cpp
@@ -471,6 +471,11 @@ void JackAlsaDriver::SetTimetAux(jack_time_t time)
     fBeginDateUst = time;
 }
 
+int JackAlsaDriver::PortSetDeviceMetadata(jack_port_id_t port_id, const char* pretty_name)
+{
+    return fEngine->PortSetDeviceMetadata(fClientControl.fRefNum, port_id, pretty_name);
+}
+
 void JackAlsaDriver::WriteOutputAux(jack_nframes_t orig_nframes, snd_pcm_sframes_t contiguous, snd_pcm_sframes_t nwritten)
 {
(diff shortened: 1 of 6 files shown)


falkTX:
Renaming the function sounds good to me.
Note that you changed `system:midi_capture_` name there to `system_midi:capture_`, which is not fine, please revert that section

iamluciano:
There was an existing naming inconsistency across the various MIDI drivers (some using `system`, some others `system_midi`), I cannot recall if I picked the convention from `coremidi` or `winmme`. If you think `system:midi_*` should be the standard I can also try to make all drivers follow that pattern. Or just revert this one.

falkTX:
If there is no consistency somewhere, create a new ticket or PR for it, do not join it in this one.
the string before the `:` represents the client name, so it cannot just be changed at will

iamluciano:
Sure, sounds reasonable. Reverted change in last commit.

iamluciano:
Think I got it... `alsarawmidi`, `coremidi` and `winmme` are implemented as separate drivers, in contrast to `raw` and `seq`, hence the need for separate client names. I noticed that _drivers_ are actually _clients_ (is this correct?), for example `JackAlsaDriver` does this:  `alsa_rawmidi_new((jack_client_t*)this)`. There is no inconsistency at all. Sorry for the confusion.

falkTX:
Thanks, merging now.

PR #510, opened 2019-10-10
Catch by reference

tartina:
Compiling with -Wcatch-value I got this warning:
../linux/alsarawmidi/JackALSARawMidiPort.cpp:112:29: warning: catching polymorphic type ‘class std::exception’ by value [-Wcatch-value=]


tartina:
Sorry wrong branch

PR #511, opened 2019-10-10
Fix warning: catching polymorphic type ‘class std::exception’ by value

tartina:
Compiling with -Wcatch-value I got this warning:
../linux/alsarawmidi/JackALSARawMidiPort.cpp:112:29: warning: catching polymorphic type ‘class std::exception’ by value [-Wcatch-value=]


falkTX:
Thanks, will merge now

PR #512, opened 2019-10-10
Fix compilation on ARM

tartina:
size_t i is needed if HAVE_EXECINFO_H is defined


falkTX:
Makes sense. though can you remove the space in `( defined` please?
I will merge after that

tartina:
Removed extra spaces

falkTX:
Thanks!

PR #520, opened 2019-10-29
v1.9.14

falkTX:


schnitzeltony:
With packager's hat on: Will you tag this as 'official release?

falkTX:
soon...

did you test the release? that would speed up things.
just waiting for confirmation on @wtay that the dbus fix is reliable.

schnitzeltony:
Ahh - ok - thanks

Up to now I just build tested. Can do a run test tonight - and drop a note.

wtay:
Hi, I can confirm that the dbus patched version works. W_

schnitzeltony:
I know it is not a systematic test but I did run it for some ours with things I use to do usually and no regressions pooped up.

PR #528, opened 2019-11-28
Fixed macosx xcode targets to use .mm extn vs .cpp for OBJC++

pteeson:
As the summary states. Most changes were to coremidi and coreaudio frameworks - 32-bit and 64-bit. Idid this by editing the target as needed. I did this using Xcode 8.3 and Xcode 9.2 on macOS Sierra 10.12.6

There is still one error when building see attached.
32-bit build


falkTX:
I dont know the reasons "aften" is required, I guess old osx versions lacked something that jack2 requires?

Regarding the commit, I am okay with merging the diff (it does not affect other things in the project), but please remove the `pteeson.xcuserdatad` folder, as that is specific config for your user

pteeson:
OK I deleted pteeson.xcuserdatad from my working copy, committed and pushed to origin.
I also added *,xcuserdatad to the root .gitignore. 
If that’s the wrong place I could move it into machos directory.
Or whatever you think is best.

The 32 and 64 bit builds tested OK on both Xcode 8.3 and 9.2.
But I’m still left with that one error wrt aften. No idea what that is about.

pteeson:
I made a commit today wrt JackAC3Encoder.cpp re-added to Driver/Audio (comments)
That got rid of the "aften” issue.

But now leaves me with this error for Xcode 8.3 and Xcode 9.2 both 32-bit & 64-bit.



Did a search and found it in the Windows folder.
No idea of the best way to proceed.

Also sorry but there are a couple of .xcuserdatad files showing up on the pull.
Can you advise me on what to do to prevent this?
Apparently putting it in the root  .gitignore didn’t work.

Sorry for my ignorance….

falkTX:
something like:
```
rm macosx/Jackdmp.xcodeproj/project.xcworkspace/xcuserdata/pteeson.xcuserdatad/UserInterfaceState.xcuserstate
git rm macosx/Jackdmp.xcodeproj/project.xcworkspace/xcuserdata/pteeson.xcuserdatad/UserInterfaceState.xcuserstate
git commit -a -m "remove unneeded files
git push
```

> But now leaves me with this error for Xcode 8.3 and Xcode 9.2 both 32-bit & 64-bit.

what error? did you try to upload a file or something? I dont see it...

pteeson:
(1) But now leaves me with this error for Xcode 8.3 and Xcode 9.2 both 32-bit & 64-bit.
> what error? did you try to upload a file or something? I dont see it...
> 
Here is what I should have attached.

 
I did find the header in the Windows folder and it has a link to the source.
Of course I can go to that source and add it into the machos folder.
But I have no idea if that is what should be done or why it is required in machos.


(2) Since then I edited the root .gitignore — see the pull details.
Replaced *.xcuserdatad with simple xcuserdata which seems to be what Mac users do.
Let me know if that worked OK

(3) Is there a way to tell if things work properly now for the pull?
There have bee 5 commits for the pull so I have no idea how it all stands right now.

Thanks for your patience…..

falkTX:
`samplerate.h` is from the libsamplerate project, one of the dependencies of the jack examples/tools.
setting that up with xcode is not something I have experience with, and out of the scope of the jack project anyway.

pteeson:
I give up and would like to close this pull request. Thank you for your help.
My intention was to use the C++ version of Jack in an app I am working on (as a hobby). And hoped to build the macOS version.

falkTX:
Sorry to hear that.
The current build system is not the best, and xcode is not something I can help with.
@dvzrv and @RedTide are trying to change it from waf to meson, which should make building jack2 much easier.

pteeson:
For cross platform apps I switched from autotools to CMake. 
That still allows each platform to use their favourite GUI or whatever.
And it seems to be the mainstream these days for large projects.
It allows a very nice abstraction of things.

Anyway I don’t have time at this point to pursue Jack2 for macOS…. Maybe later.

Thanks again for your help

PR #529, opened 2019-12-05
Fix compilation of documentation

tartina:
Custom footer without date is not needed in documentation anymore.

That was an old Fedora patch, I thinks it's not needed anymore.
If that file is not present, doc compilation fails.


falkTX:
Makes sense, thanks

PR #536, opened 2020-01-16
example-clients: Use c++ compiler for jack_simdtests

kraj:
It uses c++ sources and runtime therefore its best to use c++ compiler
to build it so it can find the correct runtime, cross compiling with
clang fails

x86_64-yoe-linux-ld: example-clients/simdtests.cpp.28.o: undefined reference to symbol '__cxa_call_unexpected@@CXXABI_1.3'

### DIFF
--- a/example-clients/wscript
+++ b/example-clients/wscript
@@ -49,10 +49,15 @@ def build(bld):
         else:
             use = ['clientlib']
 
-        if bld.env['IS_MACOSX']:
-            prog = bld(features='c cprogram', framework = ['Foundation'])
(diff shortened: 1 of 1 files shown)


falkTX:
This wont work for macOS.
I suggest to store the `features` in a string and pass that to the `bld` function.
Thanks

kraj:
@falkTX I have uploaded a new version, please review

pruperting:
FYI - this change was the only way I could compile jack2 on an Odroid N2+, niche userbase I guess, but it worked for me.

PR #540, opened 2020-01-24
Windows: Use widechar instead of char when loading drivers.

kmatheussen:
Fixes loading drivers/internals if jack is installed in a directory with special characters.


falkTX:
Patch looks clean, will merge, thanks.

PR #543, opened 2020-01-28
One should avoid gettimeofday for purposes of timestamping.

MiesSuomesta:
it is possible to mis calculate time between timestamps
if something/user changes system date. The monotonic time is
much more safer to use for timestamp purposes.

Better to let monotonic time as default timestamp time.

### DIFF
--- a/linux/JackLinuxTime.c
+++ b/linux/JackLinuxTime.c
@@ -122,7 +122,12 @@ static jack_time_t jack_get_microseconds_from_hpet (void)
 
 #endif /* HPET_SUPPORT */
 
-#define HAVE_CLOCK_GETTIME 1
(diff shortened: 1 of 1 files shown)


h1z1:
FWIW ntpd and chrony will drive you insane over this too.

imaami:
This PR doesn't remove the use of `gettimeofday()`, but actually reverts back to using it instead of `clock_gettime()`.

Fun fact: `git log -p --follow -- linux/JackLinuxTime.c` shows that the `gettimeofday()` call in JackLinuxTime has actually always been dead code in JACK2. Everything between `#ifndef HAVE_CLOCK_GETTIME` and the next `#else` should probably just be deleted.

PR #546, opened 2020-02-04
Ami generic

amiartus-adit-jv:
Hello,

these are a few minor fixes that were found as part of broader development which will be part of separate pull requests.

Adam


common/JackRequest.h
```diff
@@ -90,7 +90,7 @@ struct JackRequest
         kGetUUIDByClient = 37,
         kClientHasSessionCallback = 38,
         kComputeTotalLatencies = 39,
-        kPropertyChangeNotify = 40
+        kPropertyChangeNotify = 40,
```
falkTX: this comma is really not needed for anything.
it actually breaks compilation if you include the header in something being compiled in strict/pedantic mode.

amiartus-adit-jv: aha wasnt aware of that, the main reason for it is if you add a new enum or member, you modify only one line and resulting diff is therefore smaller and highlights only important change

falkTX: yeah, that is usually a good reason. but "strictly" speaking, it breaks pedantic language checks.

twischer-adit:
Can be closed replaced by

- https://github.com/jackaudio/jack2/pull/586
- https://github.com/jackaudio/jack2/pull/575

PR #547, opened 2020-02-05
Jack Audio Multi Client Support

amiartus-adit-jv:
Hello,

these fixes add support for Jack "multi client" mode. By multi client I mean allowing to create multiple clients who connect to different servers from one process.

In current Jack version this would result in  overwriting synchronization and signalling primitives of shared memory. Each created client would overwrite client before, so that previous client would not be able to accept process and other callbacks and notify server of graph competition.

The patch introduces "global context" reference that holds instance of global and shared memory variables for each created client. This global context reference is then passed to all objects who require it.

Following test application can be used to verify this use case:

`
#include <stdint.h>
#include <stdbool.h>
#include <jack/jack.h>
#include <iostream>
#include <signal.h>
#include <unistd.h>
#include <string>

#define CLIENTS_NUM 8
#define PORTS_NUM 2

static jack_client_t* clients[CLIENTS_NUM] = {0};

static jack_port_t* ports[CLIENTS_NUM][PORTS_NUM] = {{0}};

char servers[CLIENTS_NUM][32] = {
	"acoustic",
	"acoustic",
	"acoustic",
	"acoustic",
	"media",
	"media",
	"media",
	"media",
};

static int finished = 0;

void kill_handler(int sig)
{
	finished = 1;
}

static int backend_jack_process (const jack_nframes_t nframes, void *arg)
{
	return 0;
}

int main()
{
	jack_status_t status;
	jack_options_t options = (jack_options_t) (JackNoStartServer | JackServerName);
	const unsigned long flags = JackPortIsOutput;

	for (int i = 0; i < CLIENTS_NUM; ++i)
	{
		clients[i] = jack_client_open("client_name", options, &status, servers[i]);
		jack_set_process_callback(clients[i], backend_jack_process, nullptr);

		for (int j = 0; j < PORTS_NUM; ++j)
		{
			std::string portname = "port";
			portname += std::to_string(j);
			ports[i][j] = jack_port_register(clients[i], portname.c_str(), JACK_DEFAULT_AUDIO_TYPE, flags, 0);
		}

		jack_activate(clients[i]);
	}

	if(signal(SIGTERM, kill_handler))
		std::cout << "cant catch";

	usleep(6000000);
	
	for (int i = 0; i < CLIENTS_NUM; ++i)
	{
		jack_deactivate(clients[i]);
		jack_client_close(clients[i]);
	}

	exit(0);

	return 0;
}
`

verification of changes is done on:

    qnx platform
    aarch64 platform

Adam

### DIFF
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -291,7 +291,7 @@ static inline bool CheckBufferSize(jack_nframes_t buffer_size)
     return (buffer_size >= 1 && buffer_size <= BUFFER_SIZE_MAX);
 }
 
-static inline void WaitGraphChange()
+static inline void WaitGraphChange(JackGlobals *globals)
 {
     /*
     TLS key that is set only in RT thread, so never waits for pending
@@ -299,8 +299,8 @@ static inline void WaitGraphChange()
     */
 
     if (jack_tls_get(JackGlobals::fRealTimeThread) == NULL) {
-        JackGraphManager* manager = GetGraphManager();
-        JackEngineControl* control = GetEngineControl();
+        JackGraphManager* manager = globals->GetGraphManager();
+        JackEngineControl* control = globals->GetEngineControl();
         assert(manager);
         assert(control);
         if (manager->IsPendingChange()) {
@@ -353,7 +353,8 @@ LIB_EXPORT void* jack_port_get_buffer(jack_port_t* port, jack_nframes_t frames)
         jack_error("jack_port_get_buffer called with an incorrect port %ld", myport);
         return NULL;
     } else {
-        JackGraphManager* manager = GetGraphManager();
+        JackGlobals *global = JackGlobals::PortGlobal(port);
+        JackGraphManager* manager = global->GetGraphManager();
         return (manager ? manager->GetBuffer(myport, frames) : NULL);
     }
 }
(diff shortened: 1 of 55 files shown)


falkTX:
I am sorry, but I am not going to merge this.
You can keep the PR open for anyone interested on it, but I would not want it part of the JACK project.

Let me explain...
It is first time I see ever someone suggesting such a thing - multiple jack servers for a single process. You went even further and actually implemented the whole thing. Very impressive!
But... even for the "jack server under multiple users" scenario, that is already not very well supported (due to lack of interest or users requesting it). Your approach is even less requested.

What is the usecase here?

I see some useful commits there, I can cherry-pick those so your diff ends up smaller.

amiartus-adit-jv:
Hi Filipe, thank you for having a look. If any commits are merged that's already great, since it decreases maintenance efforts and will also make it easier in future to submit more PR with features/patches.

As for the use case, we are running multiple instances of jack server (different sample rate requirements + no re-sampling constraint). On top of that we are running a virtualization host that spawns an initialization process for guest OS, in this process we are connecting to multiple jack servers. As there is only one process spawned during the initialization of guest OS we have to make sure the static objects are correctly shared.

From architectural/requirements point of view it has to go this way for us therefore this change request came to be.

amiartus-adit-jv:
Hi Filipe, thank you for having a look. If any commits get merged thats already fine and helping us to maintian with less effort.

As for the use case, we are running multiple instances of jack server (different sample rate requirements + no re-sampling constraint + effort to reduce cpu load). On top of that we are running a virtualization host that spawns an initialization process for guest OS, in this process we are connecting to multiple jack servers. As there is only one process spawned during the initialization of guest OS we have to make sure the static objects are correctly shared.
 
From architectural/requirements point of view it has to go this way for us therefore this change request came to be.

twischer-adit:
This PR can be closed. It is replaced by https://github.com/jackaudio/jack2/pull/597

PR #548, opened 2020-02-05
Jack Audio Keep request socket stream always in sync

amiartus-adit-jv:
Hello,

rebased https://github.com/jackaudio/jack2/pull/386 on latest devel.

original request by Timo Wischer:

When the request socket stream is under high load a lot of write() calls are failing with EAGAIN. To avoid blocking of a real-time thread there is no chance to try to write again. Therefore the write of the message is aborted. With the current implementation write() will be called for each attribute of a request message. Therefore there is a high probability that a message will be written partly. But the reader expects an continuous stream and has no change to sync on an broken stream.

This PR tries to write one request message with one write() call. Therefore if a write() returns with EAGAIN the complete message will be ignored.

But in case of the write() returns with a partly write operation (e.g. only 8 of 12 bytes were written) there is still a chance to write partly messages and bring the reader out of sync. To overcome this issue I would like to retry to write as long as I am able to write some bytes and I did not get an EAGAIN (Therefore I would expect that the write() should not block a real time thread as long as we are in non-blocking mode).

But also with this implementation there is still a chance of partly written messages. Therefore the reader requires a possibility to detect partly written messages and find the start of the next message. I have added a start and stop integer to the request messages which contains "ReqS" and "ReqE".
If the reader does not detect the stop integer at the end of the message. There is a corrupted message. Therefore the reader will ignore this message and will search for the next start integer.

Adam


falkTX:
Thanks for the pull request!
Interesting changes, though I wonder if we could not have something that works in this scenario without having it do so many deep changes..

The use of `POST_PACKED_STRUCTURE_FORCED_AARCH64` is not good though,
please find a way around that or keep that a patch in your local sources.
Does JACK misbehave without it?

Also, did you check for side effects on regular desktop JACK usage?

amiartus-adit-jv:
Hi Filipe, thank you for having a look.

The request-oneshot was done by Timo, there were some issues with certain requests having tendency to fail if issues quickly. I think in general at least for me it makes sense to do one write instead of multiple small ones, because it reduces point of failure. Also there is less code as result and the code that's left is more maintainable (the later point is subjective opinion I guess).

Due to structure packing is disabled on AARCH64 (there are some patches in history that for some reason disabled it on gcc + aarch64 combination) the structures can not be used for inter-process communication. This was not seen before, I'm not sure about the reason, but I have a feeling it was working by a chance (introducing new variable broke it due to packing).

twischer-adit:
#386 includes the modifications of this PR, now.
Therefore this PR can be closed.

falkTX:
> Therefore this PR can be closed.

ok will do

PR #555, opened 2020-03-19
Fix segFault in case func returns -1

BrunoVernay:
jack_get_properties can return -1. But since size_t is unsigned, it is not <0.
It leads to a "segmentation fault (core dumped)  jack_property -l 4294967300"

Not sure if it is the best way to fix it.


falkTX:
Fixed in 70420f49919a6abddaf0f44b6d6474c09a674075, thanks for the heads up

PR #564, opened 2020-04-17
jack_control: return a proper exit status on DBus exception

jneen:
It is useful when scripting with `jack_control` to behave differently in the event of an error (invalid ALSA device, bad options, etc). This patch makes the script correctly return a non-zero exit status when there is an error.


falkTX:
Makes sense, thanks. Sorry for the delay.

PR #575, opened 2020-05-11
jack: remove unnecessary GPL include from LGPL code

twischer-adit:
in this case LGPL code is compiled into one library with GPL code, but
the include is not necessary and may cause problems in case
these files are separated into their own library

Change-Id: I9fc2499d60d8f25a8714c7c39b7b36206d5c5cf5
Signed-off-by: Adam Miartus <external.adam.miartus@de.bosch.com>


falkTX:
With the CI not working right at the moment I can't verify this properly.
So I will assume it works as you say. It targets develop branch, so we can hopefully catch build issues before a release if those happen
Thanks

PR #577, opened 2020-05-12
jack alsa: reoganize code for readability

twischer-adit:


falkTX:
small change that makes sense, why not. thanks

PR #578, opened 2020-05-12
metadata: Make JACK_METADATA_* constant

twischer-adit:
to ensure that it is not modified by any client.

const have internal linkage unless marked by extern


falkTX:
That is an odd way to go about it, why not simple use `const char* const NAME` ?

twischer-adit:
With your approach the variables are still located in read write data section (D)
```
$ nm -C --defined-only build/common/libjack.so.0.1.0 | grep JACK_METADATA_
000000000004a300 D JACK_METADATA_CONNECTED
000000000004a340 D JACK_METADATA_EVENT_TYPES
000000000004a380 D JACK_METADATA_HARDWARE
000000000004a3c0 D JACK_METADATA_ICON_LARGE
000000000004a400 D JACK_METADATA_ICON_NAME
000000000004a440 D JACK_METADATA_ICON_SMALL
000000000004a480 D JACK_METADATA_ORDER
000000000004a4c0 D JACK_METADATA_PORT_GROUP
000000000004a500 D JACK_METADATA_PRETTY_NAME
000000000004a540 D JACK_METADATA_SIGNAL_TYPE
```

With the approach of this patch it is really located in the read only data section (R):
```
$ nm -C --defined-only build/common/libjack.so.0.1.0 | grep JACK_METADATA_
000000000004a300 R JACK_METADATA_CONNECTED
000000000004a340 R JACK_METADATA_EVENT_TYPES
000000000004a380 R JACK_METADATA_HARDWARE
000000000004a3c0 R JACK_METADATA_ICON_LARGE
000000000004a400 R JACK_METADATA_ICON_NAME
000000000004a440 R JACK_METADATA_ICON_SMALL
000000000004a480 R JACK_METADATA_ORDER
000000000004a4c0 R JACK_METADATA_PORT_GROUP
000000000004a500 R JACK_METADATA_PRETTY_NAME
000000000004a540 R JACK_METADATA_SIGNAL_TYPE
```

mgeier:
Is this supposed to be a breaking change?

FYI, it breaks my Python bindings: https://github.com/spatialaudio/jackclient-python/issues/95

I can of course update my header declaration to the new style, but then it will stop working with older versions of the JACK library.

Is there a way to define this to be compatible with old and new versions?

falkTX:
It is not supposed to be a breaking change.
You are parsing the library in a weird way though, basically doing a `dir()` over all the symbols exposed..? so I expect some side effects from this, please debug further if you can.
I would say that the previous way of these symbols being exported was wrong, and now has been corrected.

mgeier:
Thanks for looking into this!

> It is not supposed to be a breaking change.

OK, good to know.

> You are parsing the library in a weird way though, basically doing a `dir()` over all the symbols exposed..? so I expect some side effects from this, please debug further if you can.

Don't worry about the `dir()` thing, that's just to assign the decoded strings to "proper" Python strings in one go.
It is not relevant to the problem, the Python interpreter also crashes if I simply do this:

```python
_lib.JACK_METADATA_ORDER[0]
```

The crash also happens if I do

```python
_ffi.string(_lib.JACK_METADATA_ORDER)
```

Note that I'm still using the old definition in `jack_build.py`:

```c
extern const char* JACK_METADATA_ORDER;
```

If I change this to the new definition, the crash goes away, but then the constants contain garbage values when using an older JACK version (but there is no crash).

> I would say that the previous way of these symbols being exported was wrong, and now has been corrected.

I don't have enough knowledge about this to have an opinion, I'm just reporting a breakage (which might as well be in my Python code or in CFFI or in Python itself or ...).

falkTX:
> ```python
> ```python
> 
> ```
> 
> 
> Note that I'm still using the old definition in jack_build.py:
> ```c
> extern const char* JACK_METADATA_ORDER;
> ```
> ```

I dont understand this part, how are you using this under python?

mgeier:
I'm using https://cffi.readthedocs.io/ to access the C API of JACK.

There are multiple "modes", I'm using "out-of-line ABI mode", which means that the C definitions are parsed by CFFI on installation, but `dlopen()` is used at runtime to link to `libjack`.

The currently used C definitions are there:
https://github.com/spatialaudio/jackclient-python/blob/3f86c7478c4e5679dc2f6f0080c7a94e31e4b350/jack_build.py#L282-L291

Those are just copied from the JACK header files (before they have been changed by this PR).

BTW, if we can't find a solution that's compatible with both old and new JACK versions, I can simply hard-code those constants in Python and ignore the C definitions from the JACK library.

PR #582, opened 2020-05-14
use feature test macro correctly

HenryHu:
The feature test macro should be defined, not tested. See [Feature Test Macros](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html):
>    Macro: _POSIX_C_SOURCE

>    Define this macro to a positive integer to control which POSIX functionality is made available. The greater the value of this macro, the more functionality is made available.
>    ....
>    If you define this macro to a value greater than or equal to 200112L, then the functionality from the 2001 edition of the POSIX standard (IEEE Standard 1003.1-2001) is made available. 

sem_timedwait is defined if _POSIX_C_SOURCE >= 200112L, so define it as such. If a system does not support this, it will be undefined and there will be a compilation error.

### DIFF
--- a/posix/JackPosixSemaphore.cpp
+++ b/posix/JackPosixSemaphore.cpp
@@ -17,6 +17,8 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
 */
 
+#define _POSIX_C_SOURCE 200112L
+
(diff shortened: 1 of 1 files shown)


falkTX:
Thanks, but this code is not used since quite a few years now.
So I will just merge, as it is safe to do so anyway.

PR #592, opened 2020-06-05
alsa: fix the poll interval

twischer-adit:
should be set in milliseconds

### DIFF
--- a/linux/alsa/alsa_driver.c
+++ b/linux/alsa/alsa_driver.c
@@ -900,7 +900,7 @@ alsa_driver_set_parameters (alsa_driver_t *driver,
 	driver->period_usecs =
 		(jack_time_t) floor ((((float) driver->frames_per_cycle) /
 				      driver->frame_rate) * 1000000.0f);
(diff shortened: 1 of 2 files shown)


kmatheussen:
Wow, what difference does this patch do?

lucianoiam:
Looks like `poll()` is getting a value in µs instead of ms somewhere in the ALSA driver code and the patch corrects that.

kmatheussen:
Yes, that's what the patch do. But what differece does it do? For the user?

kmatheussen:
(s/do/does/g)

twischer-adit:
Hello @kmatheussen,

> Wow, what difference does this patch do?

The Xrun recovery will be started earlier and therefore the audio gap is smaller (In good conditions even not audible).

When I execute JACK with a period time of 8ms and the sound device fails, I would wait 12 seconds on poll till I start the recovery.
With this patch we are only waiting 12 ms till starting the recovery.

falkTX:
FYI this is causing a regression where jack is unable to start on some systems. See #614
It is not optimal, but I am reverting the change until we figure out a proper solution.

PR #594, opened 2020-06-05
alsa: add log in case of poll was interrupted

twischer-adit:
poll may be interrupted due to signal, in that case print error log


twischer-adit:
Done

falkTX:
cool, thanks!

PR #595, opened 2020-06-05
alsa: in case of poll timeout print error

twischer-adit:
This is always an audible issue. Therefore print the corresponding error message also in non verbose mode.


twischer-adit:
Done

falkTX:
alright, that works.

PR #606, opened 2020-06-30
posix: fifo: use right prototype

twischer:
Function argument list is not matching with implementation.


falkTX:
I am curious, how did it build before without this?

falkTX:
Answering my own question.. it "worked" before simply because it is not used.
So does not hurt to fix.

Issue #395, opened 2018-11-14
jack_client_close is locking when the jack server is periodically restarts


Gravechapa:
Hi, i'm working on an application which must automatically reconnect to the jack server. But i'm run into some trouble when started to test it. After several times of restarting the jack server, my application ended up stuck at jack_client_close.
Here's example to reproduce the bug:
```
#include <jack/jack.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <pthread.h>

jack_port_t *input_port;
jack_port_t *output_port;
jack_client_t *client;
pthread_spinlock_t spinlock;
int stop = 0;

int process (jack_nframes_t nframes, void *arg) {
    jack_default_audio_sample_t *in, *out;

    in = (jack_default_audio_sample_t *)jack_port_get_buffer (input_port, nframes);
    out = (jack_default_audio_sample_t *)jack_port_get_buffer (output_port, nframes);
    memcpy (out, in,
        sizeof (jack_default_audio_sample_t) * nframes);

    return 0;   
}

void jack_shutdown(void *arg) {
    pthread_spin_lock(&spinlock);
    stop = 1;
    pthread_spin_unlock(&spinlock);
}

void  silent_jack_error_callback(const char *msg) {}

int main() {
    pthread_spin_init(&spinlock, 0);
    jack_set_error_function(silent_jack_error_callback);
    const char *client_name = "simple";
    const char *server_name = NULL;
    jack_options_t options = JackNullOption;
    jack_status_t status;
    while(true){
    client = jack_client_open (client_name, JackNoStartServer, &status, server_name);
    if (client == NULL) {
		fprintf (stderr, "jack_client_open() failed, "
			 "status = 0x%2.0x\n", status);
		if (status & JackServerFailed) {
			fprintf (stderr, "Unable to connect to JACK server\n");
		}
		continue;
	}
    
    jack_on_shutdown(client, jack_shutdown, 0);
    jack_set_process_callback (client, process, 0);

    input_port = jack_port_register (client, "input",
                                    JACK_DEFAULT_AUDIO_TYPE,
                                    JackPortIsInput, 0);
    output_port = jack_port_register (client, "output",
                                    JACK_DEFAULT_AUDIO_TYPE,
                                    JackPortIsOutput, 0);
    
    if (jack_activate (client)) {
            fprintf (stderr, "cannot activate client");
            exit (1);
    }
    while (1) {
        pthread_spin_lock(&spinlock);
[...]

PR #613, opened 2020-07-09
[RFC][WIP] posix: Set mutex type to PTHREAD_MUTEX_ERRORCHECK

fps:
When looking for a way to debug the deadlock in https://github.com/jackaudio/jack2/issues/395 I stumbled across this advisory:

https://wiki.sei.cmu.edu/confluence/display/c/POS04-C.+Avoid+using+PTHREAD_MUTEX_NORMAL+type+mutex+locks

In the end it didn't detect the deadlock (which is not a proof of it not existing) but I'll leave the PR here for your consideration. Might be helpful in the future..

### DIFF
--- a/posix/JackPosixMutex.cpp
+++ b/posix/JackPosixMutex.cpp
@@ -28,7 +28,10 @@ namespace Jack
     JackBasePosixMutex::JackBasePosixMutex(const char* name)
         :fOwner(0)
     {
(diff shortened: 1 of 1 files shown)


falkTX:
having it always on is not the most useful, do we really need it all the time?
If you wrap it around a `#ifdef BUILD_DEBUG` (so that debug builds have it, otherwise everything still the same) I will merge the PR. thanks!

PR #615, opened 2020-07-13
Improvements of MinGW/waf + build guide

HaHeho:
This is a followup to #495. I was finally able to build with waf and MinGW on Windows (for the first time). So thank you very much for these contributions. However, I had to do some modifications in order to successfully build. Some points like missing driver dependencies for a successful execution of the built binaries seem quite essential to me.

However, I have little experience with the actual build process of JACK or the dependencies. Some points should be quickly checked by somebody who has more insight. I will leave comments at the respective lines.

As discussed in #495 the README_MINGW is certainly not the greatest experience to build. However, now it should be a little bit more fail proof at least.

### DIFF
--- a/windows/README_MINGW
+++ b/windows/README_MINGW
@@ -4,6 +4,7 @@ HOW TO COMPILE JACK USING WAF AND MINGW
 Rev. 1 - 2019-09-01 - First version
 Rev. 2 - 2019-09-18 - Include contributions from @Schroedingers-Cat
 Rev. 3 - 2019-12-14 - Include contributions from @Schroedingers-Cat
+Rev. 4 - 2020-07-13 - Include contributions from @HaHeho
 
 
 
@@ -141,7 +142,7 @@ mv /tmp/asiosdk_{version}_{date} /opt/asiosdk
 The description file in portaudio (MINGW-packages/mingw-w64-portaudio/PKGBUILD)
 needs to be patched so configure is called with the necessary flags for ASIO:
 
---with-asiodir=/opt/asiosdk   <- new option     
+--with-asiodir=/opt/asiosdk   <- new option
 --with-winapi=wmme,directx,wasapi,vdmks,asio   <- append 'asio' to existing
 
 Both static and shared library versions of portaudio are built. To prevent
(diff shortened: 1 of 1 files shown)


windows/README_MINGW
```diff
@@ -249,6 +268,11 @@ cp C:\msys64\mingw64\bin\libgcc_s_sjlj-1.dll `
   , C:\msys64\mingw64\bin\libwinpthread-1.dll `
   C:\Windows
 
+- Copy all dll files from C:\msys64\opt\jack\lib\jack to C:\Windows\jack, these
+are dependencies for libjack:
+
+cp C:\msys64\opt\jack\lib\jack\*.dll C:\Windows\jack
```
HaHeho, line 286: This was missing in the original contribution. When following all other copy / paste operations I received
```
Drivers/internals found in : C:\WINDOWS
Drivers not found in "C:\WINDOWS\jack": "C:\WINDOWS\jack/*.dll"
Driver folder not found
Could not find any drivers in driver directory!
Failed to create server object
```

After placing all build DLLs in the relative jack path under `C:\Windows`  I could run the server successfully. However, I don't know if this is the best solution or all of the DLLs are really required. Here is a list:
`audioadapter.dll, inprocess.dll, jack_dummy.dll, jack_loopback.dll, jack_net.dll, jack_netone.dll, jack_portaudio.dll, jack_proxy.dll, jack_winmme.dll, netadapter.dll, netmanager.dll, profiler.dll`

falkTX:
Just so you are aware, automated builds for jack are working now.
They are tagged and uploaded at https://github.com/jackaudio/jack2-releases

I have made a few changes in the wscript(s) in order to get this to work, so you are going to need to adjust the readme a bit.
The automated builds have jack installed in roughly the same place as the old installer (except the "program files" is now the correct one).
There is also a registry key added so others can find jackd.exe
See https://github.com/jackaudio/jack2/blob/b3603d6fb3afa201ae251ae0826bad627cbcdef4/windows/inno/win64.iss to check where all files go.
The instructions should follow this, and once they do, jack is able to start.

The updated instructions look nice overall, just take into account the updated locations and we should be good, right?

PR #626, opened 2020-08-14
Fix jack_control handling of dbus bytes

joshuarubin:
Prior to this, `jack_control dps dither s` would return an error.


falkTX:
Makes sense I guess, thanks!

PR #641, opened 2020-10-07
Clarification about jack_port_get_latency_range().

automatthias:
jack_port_get_latency_range only returns meaningful values after ports get
connected, and that is signalled via the latency callback. Saying that it's
normally used in callbacks is too soft, the docs should make it clear that the
function is not very useful outside of the callback, because you don't know
whether the port is connected / whether the latency values changed.


falkTX:
See https://github.com/jackaudio/headers/pull/9
Thanks

falkTX:
Please note it is preferable to have PRs target develop branch instead of master. But since this is a comment change only, it will not affect anything else.

PR #644, opened 2020-10-14
fix a deadlock issue in midi_dump

Krasjet:
The main loop might be blocked by `data_ready` when JACK server dies
while the program is running.

To reproduce the problem:

1. Start JACK server
2. Run `jack_midi_dump`
3. Stop JACK server
4. Press Ctrl-C in `jack_midi_dump` to send SIGINT.
5. Observe that the program doesn't stop and stuck in the main loop.

This patch forces the `data_ready` to be signaled in the signal handler
to prevent indefinite waiting.

### DIFF
--- a/tools/midi_dump.c
+++ b/tools/midi_dump.c
@@ -112,6 +112,11 @@ process (jack_nframes_t frames, void* arg)
 static void wearedone(int sig) {
 	fprintf(stderr, "Shutting down\n");
 	keeprunning = 0;
+	/* main loop might be blocked by data_ready when jack server dies. */
+	if (pthread_mutex_trylock (&msg_thread_lock) == 0) {
+		pthread_cond_signal (&data_ready);
+		pthread_mutex_unlock (&msg_thread_lock);
+	}
 }
 
 static void usage (int status) {


falkTX:
Thanks. Today was the day set for the new release, I do not have much time to test and verify this, so let's push that for the one after.
It is related to something that needs fixing - server restart deleting client-side data - so I prefer to look into this all together.

Krasjet:
I think this problem has nothing to do with JACK server, though.

The issue is caused by the `pthread_cond_wait` at the end of main loop.

https://github.com/jackaudio/jack2/blob/10eb69fd02b633383c98710269b4d09ef1667a76/tools/midi_dump.c#L227-L231

After the server dies and the `process` callback is no longer active, the `pthread_cond_signal` in `process` stops to be called

https://github.com/jackaudio/jack2/blob/10eb69fd02b633383c98710269b4d09ef1667a76/tools/midi_dump.c#L104-L107

and the main loop is stuck at `pthread_cond_wait` above. The `keeprunning` flag won't be checked until something signals the condition variable.

https://github.com/jackaudio/jack2/blob/10eb69fd02b633383c98710269b4d09ef1667a76/tools/midi_dump.c#L201

The proposed fix simply signals the condition variable in the signal handler to let the main loop move forward into the next iteration.

PR #665, opened 2020-11-07
Change returncodes of jackd on all successes to 0

### LINKED ISSUE #663: --help and --version return status code -1
~I am currently preparing to add JACK2 to the official `homebrew-core` tap for OSX.~ https://github.com/Homebrew/homebrew-core/pull/63082
A simple test to check the reported version string using `jackd --version` fails since `print_version()` always returns [`-1`](https://github.com/jackaudio/jack2/blob/v1.9.16/common/Jackdmp.cpp#L256).
Same thing goes for `jackd --help` as it is treated as a [server failure](https://github.com/jackaudio/jack2/blob/v1.9.16/common/Jackdmp.cpp#L504) and returns the default return code [`-1`](https://github.com/jackaudio/jack2/blob/v1.9.16/common/Jackdmp.cpp#L325).

If you don't have any objections, I will change both cases to `0` since they are no failures.

C0nsultant:
`--help` and `--version` do not result in server failures.
Instead of returning the default `-1`, return a more reasonable `0`.

Closes #663.

### DIFF
--- a/common/Jackdmp.cpp
+++ b/common/Jackdmp.cpp
@@ -253,8 +253,6 @@ void print_version()
     printf( "jackdmp version " VERSION " tmpdir "
             jack_server_dir " protocol %d" "\n",
             JACK_PROTOCOL_VERSION);
-    exit(-1);
-
 }
 
 int main(int argc, char** argv)
@@ -269,6 +267,7 @@ int main(int argc, char** argv)
     for(int a = 1; a < argc; ++a) {
         if( !strcmp(argv[a], "--version") || !strcmp(argv[a], "-V") ) {
             print_version();
+            return 0;
         }
     }
     const char *options = "-d:X:I:P:uvshrRL:STFl:t:mn:p:C:"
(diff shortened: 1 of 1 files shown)


common/Jackdmp.cpp
```diff
@@ -253,8 +253,6 @@ void print_version()
     printf( "jackdmp version " VERSION " tmpdir "
             jack_server_dir " protocol %d" "\n",
             JACK_PROTOCOL_VERSION);
-    exit(-1);
```
C0nsultant, line 256: As all other functions that just print also do not exit the program, I moved this logic down to the `main()` body.

common/Jackdmp.cpp
```diff
@@ -497,10 +496,11 @@ int main(int argc, char** argv)
 
             default:
                 fprintf(stderr, "unknown option character %c\n", optopt);
-                /*fallthru*/
+                goto destroy_server;
```
falkTX, line 499: this is changing previous behaviour.
please add printing of usage. so it works just like before.

C0nsultant: Fixed. Since there is no more fall-through I also moved `default:` to the bottom.

falkTX:
Looks good, thanks!

PR #670, opened 2020-11-16
Use SYS_futex instead of __NR_futex

kraj:
SYS_futex is expected from system C library.
in glibc (/usr/include/bits/syscall.h defines it in terms of of NR_futex)
rv32 is using 64bit time_t from get go unlike other 32bit architectures
in glibc, therefore it wont have NR_futex defined but just NR_futex_time64
this aliases it to NR_futex so that SYS_futex is then defined for rv32

Upstream-Status: Pending


falkTX:
I have not seen the need for NR_futex_time64 yet myself, but this change still keeps the build working, so seems good to me.
Thanks!

PR #673, opened 2020-11-27
Update README_MINGW to reflect 2020 changes

lucianoiam:
- Make it clear it is not the official way to build JACK *for* Windows, but a way to build JACK *on* Windows.
- Update build instructions for JackRouter. These could be copied/moved to the jack-router repo at some stage. Since it is a small VS project there is not really much to explain, though.


falkTX:
Thanks a lot, makes perfect sense.

I will try to get JackRouter building with mingw sometime soon, not sure if I will have time on weekend..
But I am sure you will notice once that happens.

PR #692, opened 2020-12-28
Fixing macOS version API use

robertchin:
The compiler definition was incorrect. MAC_OS_X_VERSION_10_5 is always defined on MacOS 10.5 and higher. We need to check for the currently platform version.

### DIFF
--- a/macosx/coreaudio/JackCoreAudioAdapter.mm
+++ b/macosx/coreaudio/JackCoreAudioAdapter.mm
@@ -911,21 +911,21 @@ int JackCoreAudioAdapter::OpenAUHAL(bool capturing,
     }
 
     // AUHAL
-#ifdef MAC_OS_X_VERSION_10_5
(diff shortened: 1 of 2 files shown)


robertchin:
FYI the old definition caused the source base to take the older deprecated component manager codepath rather than the new one (that already existed). Likely on macOS 11 or possibly Apple Silicon these deprecated APIs were turned into hard failure if they were compiled on or after macOS 11.

falkTX:
that makes sense, thanks.
the check should be `>= MAC_OS_X_VERSION_10_5` though, no?

robertchin:
AudioComponentInstanceNew is available starting 10.6. So technically it should be >= MAC_OS_X_VERSION_10_6. But I tried to keep the existing check mostly the same --previously it was #ifdef MAC_OS_X_VERSION_10_5 for the component manager code.

PR #694, opened 2021-01-02
Fix compilation of documentation

tartina:
Custom footer without date is not needed anymore


tartina:
Somehow you inserted again the old Fedora patch, which is not used anymore

falkTX:
Thanks!

Issue #652, opened 2020-10-20
jack_control doesn't run multiple commands


dj-foxxy:
Our system that starts jack generates this;

```jack_control ds alsa dps device hw:K6 dpr capture dpr playback dps rate 44100 dps period 128 dps nperiods 2 dpr hwmon dpr hwmeter dpr duplex dpr softmode dpr monitor dpr dither dpr inchannels dpr outchannels dpr shorts dpr input-latency dpr output-latency dpr midi-driver epr name epr realtime epr realtime-priority epr temporary epr verbose epr client-timeout epr clock-source epr port-max epr replace-registry epr sync epr self-connect-mode epr slave-drivers```

`jack_control` used to run each subcommand, however, last night I noticed it would only run the first one then exit.

I'm I doing something wrong or is this a bug?

falkTX:
Likely a regression.

PR #695, opened 2021-01-03
Fix bug stopping jack_control doing many ops per execution

dj-foxxy:
This bug was introduced here: https://github.com/jackaudio/jack2/commit/5d80e06c08019aaceca01fa34f0cfa3b3fb73653#diff-7d08f37916cbe03a6368b218295081808a648cfe8d849d5b29326eb9228ed501R400

We use this script loads. Could I become the "maintainer" of it? If so, what the versions of Python does jack2 support?

Fixes issue: https://github.com/jackaudio/jack2/issues/652

Thanks, Pete.

### DIFF
--- a/tools/jack_control
+++ b/tools/jack_control
@@ -397,7 +397,7 @@ def main():
             print("DBus exception: %s" % str(e))
             return 1
 
-        return 0
+    return 0
 
 if __name__ == '__main__':
     sys.exit(main())


falkTX:
Good catch!

If you have any fixes, just send them in.
But it is best to not change this file a whole bunch now, unless there is a very good reason to do so.

dj-foxxy:
I'd like to do some refactoring so that is bug (or type of bug)  can't happy again. The bug happened because of the long while/try/if-elif nested block, which makes it hard to know which block a line of code is in. I want to move each command handler into its own function.

falkTX:
That makes sense, and will be welcome yes.
Create another pull-request against develop branch again when ready, thank you.

I will push your previous fix to v1.9.17 which is going out in 9 days, but then this new change would go into v1.9.18 in order to give time for review and testing.

dj-foxxy:
Cool. What versions of Python does this project support?

falkTX:
python2 is EOL, so dont bother with that.
python3.4 should be the minimum target.

PR #700, opened 2021-01-07
Move jack_control command handlers to own functions

dj-foxxy:
I've tried to change as little as possible within the handlers. I've switched our setup to use this version so will be testing it.

### DIFF
--- a/tools/jack_control
+++ b/tools/jack_control
@@ -151,253 +151,349 @@ def maybe_print_param_constraint(iface, param):
         for value in values:
             print(("%*s'%s' - %s" % (1 + max_len - len(str(value[0])), "", str(value[0]), str(value[1]))))
 
+
+class Context:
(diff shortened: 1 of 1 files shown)


nedko:
This commit breaks python2 compatibility of jack_control

While I like the intent, in LADI/jackdbus I'm playing conservative and keeping python2 compatibility, for its stable interface (even if it has other issues) and thus the non-deployability to python2 systems is a concern.

falkTX:
main jack2 doesnt care about python2, any recent enough distribution does not provide it as it is unmaintained by a very long time and is now a security risk.

there was some initial work to move from waf to meson, which will help with this.

dj-foxxy:
Hi all. We've moved away from JACK so I'm unlikely to work on this PR. Feel free to close. :)

PR #741, opened 2021-03-29
JackLinuxFutex: treat maximum TimedWait() as Wait() and retry on EINTR

chzchzchz:
FUTEX_WAIT may be interrupted by a signal and return EINTR. If a client
process takes a signal while on the futex, the jack client may error out
with no way to recover despite the signal being safe. Instead, retry if
errno is set to EINTR.

### DIFF
--- a/common/JackClient.cpp
+++ b/common/JackClient.cpp
@@ -31,6 +31,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 #include <math.h>
 #include <string>
 #include <algorithm>
+#include <climits>
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


chzchzchz:
jack2 + golang hits this path for me after a few hundred process callbacks usually. The golang runtime fires off signals for whatever it does and this causes the waiting futex to fail. NB. TimedWait will still have trouble with EINTR; I left it alone because the timeout must be manually updated to account for the elapsed time and I've only seen the problem on the usec=0x7f...f path so far.

falkTX:
Looks good, thanks.
I will give this some testing on my side before merging.

falkTX:
Ok this seems fine.
I didn't realize that I had the very same problem in carla, which was fixed in the same way.
I will merge this and adapt the same to the timedwait call too.
Thanks!

PR #752, opened 2021-04-22
__STDC_FORMAT_MACROS needs to be first

scopatz:
With some compilers and platforms `__STDC_FORMAT_MACROS` needs to be defined before anything else, as stated in the notes of the C++ refence: https://en.cppreference.com/w/cpp/types/integer#Notes. 

I was seeing issues on gcc like those of #408 and https://stackoverflow.com/questions/52715250/strange-compiler-behavior-with-priu64-in-c

### DIFF
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -18,6 +18,8 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
 */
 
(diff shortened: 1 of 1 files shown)


imaami (approved):
(Disclaimer: did not actually test this in any way, but...)

From past experience I can confirm that the reordering of this macro is indeed a necessary thing.

falkTX:
Why not, thanks!

Issue #749, opened 2021-04-18
jack_transport_locate crashes on macOS / Apple Silicon / HomeBrew


cme:
## Describe the bug

Use of the `jack_transport_locate` function seems to reliably cause a crash in clients on Apple Silicon, with the current Homebrew packages of JACK.

This can be seen just be running `jack_test -v`, the process will crash after calling `jack_transport_locate()`

## Environment

* **JACK Version:** 1.9.17 (current Homebrew version)
* **Operating System:** macOS Big Sur 11.2.3, Apple Silicon M1
* **Installation:** Homebrew

## Steps To Reproduce

```bash
% jack_test -v
```

### Expected vs. actual behavior

Expected behaviour (from a Linux ARM 32-bit host):

```
Transport is rolling...
Test jack_transport_locate while rolling
Transport current frame = 4096
Transport is rolling...
Do jack_transport_locate
Locating.... frame = 5120
Transport is rolling...
Locating.... frame = 5120
Transport is rolling...
Locating.... frame = 6144
Transport is rolling...
Locating.... frame = 6144
Transport is rolling...
sync callback : Releasing status : now ready...
```

Observed behaviour:

```
Transport is rolling...
Test jack_transport_locate while rolling
Transport current frame = 6144
Transport is rolling...
Do jack_transport_locate
zsh: bus error  jack_test -v
```

falkTX:
Can you get a (back)trace?

cme:
The crash is in `CAS` (in `macosx/JackAtomic_os.h` ) under `JackTransportEngine::GenerateUniqueID()`, and the faulting instruction is a CASAL instruction (generated by the `__sync_bool_compare_and_swap` intrinsic added in #685 )

The memory address looks valid and can be read and written via the debugger, although it's in shared memory which should *probably* be mapped with `MAP_HASSEMAPHORE` but isn't. I suspected this might be a problem, but adding the flag doesn't seem to help. 

I can't seem to find any reference that mentions any new restrictions or requirements on semaphores in shared memory with these instrinsics or instructions on AArch64 macOS.

falkTX:
the code was crashing in a similar path before.
and was fixed by offseting some struct members so memory was aligned to 32bit (or 64bits not sure now)

this might need something like that too.

falkTX:
Odd enough, using something like https://kx.studio/Applications:Carla I remember the transport controls behaving normal, no crashes from that.

cme:
[...]

PR #761, opened 2021-05-30
Fix alignment of fields for atomic accesses

cme:
Fix for issue #749.

Atomic accesses on AArch64 require atomically accessed fields to be naturally aligned. The packed structure layouts used on macOS to ensure than x86 and ARM processes can share memory breaks this constraint in some instances.

This change
  1. Introduces alignment checks that ensure atomically-accessed fields are aligned within their parent structure (and that these parent structures are aligned within any struct containing them). These checks are static and have no runtime overhead.
  2. Changes order of some fields, and adds some padding in one instance, to meet the alignment constraints.

### DIFF
--- a/common/JackActivationCount.h
+++ b/common/JackActivationCount.h
@@ -39,7 +39,7 @@ class JackActivationCount
 
     private:
 
-        SInt32 fValue;
+        alignas(SInt32) SInt32 fValue;
         SInt32 fCount;
 
     public:
--- a/common/JackActivationCount.h
+++ b/common/JackActivationCount.h
@@ -45,7 +45,10 @@ class JackActivationCount
     public:
 
         JackActivationCount(): fValue(0), fCount(0)
-        {}
+        {
+            static_assert(offsetof(JackActivationCount, fValue) % sizeof(fValue) == 0,
+                          "fValue must be aligned within JackActivationCount");
+        }
 
         bool Signal(JackSynchro* synchro, JackClientControl* control);
 
--- a/common/JackAtomicArrayState.h
+++ b/common/JackAtomicArrayState.h
@@ -23,6 +23,7 @@
 #include "JackAtomic.h"
 #include "JackCompilerDeps.h"
 #include <string.h> // for memcpy
+#include <cstddef>
 
 namespace Jack
 {
@@ -159,6 +160,8 @@ class JackAtomicArrayState
 
         JackAtomicArrayState()
         {
+            static_assert(offsetof(JackAtomicArrayState, fCounter) % sizeof(fCounter) == 0,
+                          "fCounter must be aligned within JackAtomicArrayState");
             Counter1(fCounter) = 0;
         }
 
--- a/common/JackAtomicArrayState.h
+++ b/common/JackAtomicArrayState.h
@@ -122,7 +122,7 @@ class JackAtomicArrayState
         // fState[2] ==> request
 
         T fState[3];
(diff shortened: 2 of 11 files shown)


miartad:
could alignas be used as well to make this more robust and explicit?

cme:
> could alignas be used as well to make this more robust and explicit?

I think the same thing as `__attribute__((aligned()))` applies to `alignas` -- having the containing struct packed forces alignment of 1 on the struct as a whole and hence all the members too.

Unless there's a usage of `alignas` I'm not aware of?

Thanks!

cme:
> > could alignas be used as well to make this more robust and explicit?
> 
> I think the same thing as `__attribute__((aligned()))` applies to `alignas` -- having the containing struct packed forces alignment of 1 on the struct as a whole and hence all the members too.

This seems to be true, but also, I now realise that these would actually work so long as instances included in containing structs are also declared as aligned. This would make the static asserts here redundant and document the intention explicitly.

This sounds better.

cme:
Now using `alignas` to set the alignment explicitly, and retaining the `static_assert`s as an additional check on the alignment of the hierarchy.

falkTX:
Let's merge this and give it a try for 1 release.
If it works fine on all systems, we might be able to re-enable struct packing for linux arm architectures too.

0EVSG:
This reportedly breaks build with Clang on PowerPC (but not with GCC on PowerPC):
```
    In file included from ../common/JackDebugClient.cpp:21:
    ../common/JackEngineControl.h:67:5: error: requested alignment is less than minimum alignment of 8 for type 'Jack::JackTransportEngine'
        alignas(UInt32) JackTransportEngine fTransport;
        ^
    ../common/JackEngineControl.h:89:5: error: requested alignment is less than minimum alignment of 8 for type 'Jack::JackFrameTimer'
        alignas(UInt32) JackFrameTimer fFrameTimer;
        ^
```

While I do not understand what causes Clang to require 8 byte alignment for these packed classes, it's probably not against the standard. Is there a reason not to use e.g.
```
alignas(JackTransportEngine) JackTransportEngine fTransport;
```
here?

cme:
Ah, there's a semantic difference between `alignas` and `__attribute__((aligned))` that I hadn't noticed. The attribute sets the minimum alignment requirement but `alignas` makes it explicitly ill-formed to set a lower alignment than the natural alignment. Changing it as you suggest is probably a good idea, yes.

I wonder why clang for AArch64 didn't complain.

0EVSG:
Funny thing, 10 minutes before your answer I get a [bug report](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261508) of this failing also on FreeBSD / Aarch64 - could you provide me the compiler version and build flags of a successful build?

Regarding correctness, I think that `alignas(JackTransportEngine)` would be the right approach for single platforms. But if you mix x86_64 and Arm64 you'd actually want the maximum alignment of both platforms. So that may be a good reason to use fixed alignment, we'd have to verify.

Microsoft seems to have a similar translation technology BTW, but I've never seen it in the wild.

0EVSG:
Ok, did an experiment with `alignas(JackTransportEngine)` [here](https://github.com/0EVSG/jack2/tree/alignas_portable), and it breaks CI for MacOS / GCC. GCC does not propagate the alignment requirements of JackTransportEngine correctly. The (presumably newer) GCC on Linux works fine, though.

So I leave that be at the moment.

PR #770, opened 2021-07-05
Add jack_position_t::tick_double, and flags around it

falkTX:
This PR adds `tick_double` to `jack_position_t` and `JackTickDouble` as a validation flag for it.
Since older versions of JACK do not expose this variable, the macro JACK_TICK_DOUBLE is provided, which can be used as build-time detection.

Example code for transport-listening clients:
```
double tick;
#ifdef JACK_TICK_DOUBLE
if (pos.valid & JackTickDouble)
    tick = pos.tick_double;
else
#endif
    tick = pos.tick;
```

Rationale:
When using JACK transport to sync between clients with precise timing requirements (such as MIDI sequencers) rounding errors would accumulate and eventually make the separate clients out of sync.
This was observed in Carla and mod-host, which use audio plugins as JACK clients. Some MIDI plugins could miss notes due to rounding errors. This change has been deployed in MOD Devices for a couple of releases already and it is known to work (that is, it corrects the situation).
Since the transport API has padding members available for use and it has been unchanged for several years (so there won't be a need to add more fields in the short or middle term), let's just go for it.

### DIFF
--- a/common/jack/types.h
+++ b/common/jack/types.h
@@ -538,17 +538,21 @@ typedef uint64_t jack_unique_t;         /**< Unique ID (opaque) */
  */
 typedef enum {
 
-    JackPositionBBT = 0x10,     /**< Bar, Beat, Tick */
-    JackPositionTimecode = 0x20,        /**< External timecode */
-    JackBBTFrameOffset =      0x40,     /**< Frame offset of BBT information */
-    JackAudioVideoRatio =     0x80, /**< audio frames per video frame */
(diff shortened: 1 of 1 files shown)


falkTX:
To add some extra notes here:

There was discussion on IRC about this potentially be unnecessary, that clients can just use `bar_start_tick` to store the non-integer part of the tick.
While the idea could work in theory, supporting it turns out to be non-trivial and from all applications that I have tested none implemented this part correctly.
Some applications do not set `bar_start_tick` at all, even though they can be run as transport master.

PR #785, opened 2021-08-04
macOS: pass JackMachSemaphore send right directly via shm

### LINKED ISSUE #784: macOS: clients can only connect 98 times per boot
## Describe the bug

On macOS
If a client with a given name connects to a server with a given name 98 times
Further attempts to connect a client with that name to a server with that name will fail until the user's session is restarted (i.e. the operating system is rebooted).

## Environment

* **JACK Version:** Observed in:
`jackdmp version 1.9.18`
`jackdmp version 1.9.19`
`development` branch as of ef2af6821
I believe bug introduced in 2a4c9f74

* **Operating System:**
```
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H15
```

```
ProductName:    macOS
ProductVersion: 11.4
BuildVersion:   20F71
```

* **Installation:** Built from source, also official jackd binaries 1.9.{18,19} from jackaudio.org

## Steps To Reproduce

```bash
# On macOS:
jackd -vd dummy &
for i in {1..100} ; do echo "TRIAL $i" ; jack_lsp || echo "FAILURE" ; done
```

**Expected**: there is no limit to the number of times a client can successfully connect to jackd before the operating system must be rebooted.
**Actual**: When a client with a given name connects to a server of a given name for the 99th (or later) time, the following output is observed. Rebooting the operating system resolves the issue.

<details>
<summary>Server output</summary>

```
% jackd -vd dummy
jackdmp 1.9.20
Copyright 2004-2016 Grame.
Copyright 2016-2021 Filipe Coelho.
Jack: JackPosixThread::StartImp : create non RT thread
Jack: JackPosixThread::ThreadHandler : start
Jack: JackDriver::Open capture_driver_name = dummy
Jack: JackDriver::Open playback_driver_name = dummy
Jack: Check protocol client = 9 server = 9
Jack: JackEngine::ClientInternalOpen: name = system
Jack: JackEngine::AllocateRefNum ref = 0
Jack: JackMachSemaphore::Allocate name = js501.system
Jack: JackEngine::NotifyAddClient: name = system
Jack: JackGraphManager::SetBufferSize size = 1024
Jack: JackConnectionManager::DirectConnect first: ref1 = 0 ref2 = 0
Jack: JackGraphManager::ConnectRefNum cur_index = 0 ref1 = 0 ref2 = 0
Jack: JackDriver::SetupDriverSync driver sem in flush mode
Jack: JackSocketServerChannel::Open
Jack: JackServerSocket::Bind : addr.sun_path /tmp/jack_default_501_0
Jack: JackSocketServerChannel::BuildPoolTable size = 1
Jack: JackEngine::Open
Jack: JackClientSocket::Connect : addr.sun_path /tmp/jack_default_501_0
Jack: JackEngine::ClientInternalOpen: name = freewheel
Jack: JackEngine::AllocateRefNum ref = 1
Jack: JackMachSemaphore::Allocate name = js501.freewheel
Jack: JackEngine::NotifyAddClient: name = freewheel
Jack: JackDriver::ClientNotify ref = 1 driver = system name = freewheel notify = 0
Jack: JackDriver::ClientNotify ref = 0 driver = freewheel name = system notify = 0
Jack: JackConnectionManager::DirectConnect first: ref1 = 1 ref2 = 1
Jack: JackGraphManager::ConnectRefNum cur_index = 0 ref1 = 1 ref2 = 1
Jack: JackDriver::SetupDriverSync driver sem in flush mode
Jack: JackAudioDriver::Attach fBufferSize = 1024 fSampleRate = 48000
Jack: JackEngine::PortRegister ref = 0 name = system:capture_1 type = 32 bit float mono audio flags = 22 buffer_size = 1024
Jack: JackGraphManager::AllocatePortAux port_index = 1 name = system:capture_1 type = 32 bit float mono audio
Jack: JackConnectionManager::AddOutputPort ref = 0 port = 1
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackAudioDriver::Attach fCapturePortList[i] port_index = 1
Jack: JackEngine::PortRegister ref = 0 name = system:capture_2 type = 32 bit float mono audio flags = 22 buffer_size = 1024
Jack: JackGraphManager::AllocatePortAux port_index = 2 name = system:capture_2 type = 32 bit float mono audio
Jack: JackConnectionManager::AddOutputPort ref = 0 port = 2
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackAudioDriver::Attach fCapturePortList[i] port_index = 2
Jack: JackEngine::PortRegister ref = 0 name = system:playback_1 type = 32 bit float mono audio flags = 21 buffer_size = 1024
Jack: JackGraphManager::AllocatePortAux port_index = 3 name = system:playback_1 type = 32 bit float mono audio
Jack: JackConnectionManager::AddInputPort ref = 0 port = 3
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackAudioDriver::Attach fPlaybackPortList[i] port_index = 3
Jack: JackEngine::PortRegister ref = 0 name = system:playback_2 type = 32 bit float mono audio flags = 21 buffer_size = 1024
Jack: JackGraphManager::AllocatePortAux port_index = 4 name = system:playback_2 type = 32 bit float mono audio
Jack: JackConnectionManager::AddInputPort ref = 0 port = 4
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackEngine::ClientNotify: no callback for notification = 9
Jack: JackAudioDriver::Attach fPlaybackPortList[i] port_index = 4
Jack: JackServer::Start
Jack: JackThreadedDriver::Start
Jack: JackPosixThread::StartImp : create non RT thread
Jack: JackPosixThread::ThreadHandler : start
Jack: JackThreadedDriver::Init real-time
Jack: JackMachThread::AcquireSelfRealTime fPeriod = 21333 fComputation = 100 fConstraint = 21333
Jack: JackMachThread::thread_policy_set res = 0
Jack: JackPosixThread::StartImp : create non RT thread
Jack: JackPosixThread::ThreadHandler : start
Jack: JackSocketServerChannel::ClientCreate socket
Jack: JackSocketServerChannel::BuildPoolTable size = 2
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 1 fd = 11
Jack: JackRequest::Notification
Jack: JackDriver::ClientNotify ref = 1 driver = freewheel name = freewheel notify = 18
Jack: JackDriver::ClientNotify ref = 1 driver = freewheel name = freewheel notify = 18
Jack: JackDriver::ClientNotify ref = 1 driver = freewheel name = freewheel notify = 18
Jack: JackDriver::ClientNotify ref = 1 driver = freewheel name = freewheel notify = 18
Jack: JackEngine::ClientNotify: no callback for notification = 4
Jack: JackEngine::ClientNotify: no callback for notification = 4
Jack: JackSocketServerChannel::ClientCreate socket
Jack: JackSocketServerChannel::BuildPoolTable size = 3
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 1 fd = 11
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 2 fd = 12
Jack: JackSocketServerChannel::Execute : poll client error err = Undefined error: 0
Jack: JackSocketServerChannel::ClientKill ref = -1 fd = 12
Jack: Client was not opened : probably correspond to server_check
Jack: JackClientSocket::Close
Jack: JackSocketServerChannel::BuildPoolTable size = 2
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 1 fd = 11
Jack: JackSocketServerChannel::ClientCreate socket
Jack: JackSocketServerChannel::BuildPoolTable size = 3
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 1 fd = 11
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 2 fd = 12
Jack: JackRequest::ClientCheck
Jack: Check protocol client = 9 server = 9
Jack: JackRequest::ClientOpen
Jack: JackEngine::ClientExternalOpen: uuid = 3, name = lsp
Jack: JackEngine::AllocateRefNum ref = 2
Jack: JackMachSemaphore::Allocate name = js501.lsp
Cannot allocate synchro
Jack: JackMachSemaphore::Destroy name = js501.lsp
Jack: JackExternalClient::Close
Jack: JackSocketNotifyChannel::Close
Jack: JackClientSocket::Close
Cannot create new client
Jack: JackSocketServerChannel::Execute : poll client error err = Socket is not connected
Jack: JackSocketServerChannel::ClientKill ref = -1 fd = 12
Jack: Client was not opened : probably correspond to server_check
Jack: JackClientSocket::Close
Jack: JackSocketServerChannel::BuildPoolTable size = 2
Jack: JackSocketServerChannel::BuildPoolTable fSocketTable i = 1 fd = 11
```

</details>

<details>
<summary>Client output</summary>

```
% jack_lsp 
Cannot open lsp client
Error: cannot connect to JACK, jack_client_open() failed, status = 0x21
```

</details>

## Analysis
I believe this happens because `JackMachSemaphore` communicates the mach port for the semaphore to the client using a dynamically generated system-global resource, a named service created via `bootstrap_register`. This is done by `JackMachSemaphore::recursiveBootstrapRegister(...)`.

The `recursive` is needed because the bootstrap server will refuse to allow us to re-register a previously-used service name (and also will not allow us to re-use a previously-created service either, I believe as our bootstrap port is not privileged and for some reason launchd thinks the dead service ports are still active?).

As such, in the event of most kinds of failure, we simply try again, appending an incrementing integer to the end of the service name. There does not seem to be any means available to us that I can find of causing the bootstrap server to forget about a previously-registered service, meaning that once registered, these service registrations persist until the user's session is restarted (i.e. reboot). We can verify this by running `jackd`, executing `jack_lsp` a few times, and inspecting the list of registered services:

<details>
<summary>launchctl print output showing registered service ports persisting</summary>

```
% launchctl print user/`id -u`
<snip>
        unmanaged processes = {
                com.apple.xpc.launchd.unmanaged.jackd.496 = {
                        active count = 13
                        dynamic endpoints = {
                                "js501.lsp-10" = {
                                        port = 0x7582b
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-9" = {
                                        port = 0x7a01b
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-8" = {
                                        port = 0x7c037
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-7" = {
                                        port = 0x79f1f
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-6" = {
                                        port = 0x7f85f
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-5" = {
                                        port = 0x7a3fb
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-4" = {
                                        port = 0x7a23b
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-3" = {
                                        port = 0x801bb
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp-2" = {
                                        port = 0x79b17
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.lsp" = {
                                        port = 0x79c1b
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.freewheel" = {
                                        port = 0x75907
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                                "js501.system" = {
                                        port = 0x159f
                                        active = 1
                                        managed = 0
                                        reset = 0
                                        hide = 0
                                }
                        }
<snip>
                }
        }
<snip>
```

</details>

Unfortunately, `recursiveBootstrapRegister(...)` has a [recursion limit](https://github.com/jackaudio/jack2/blob/852ba2b8ebc944865e75d8eb5d4550d408772f3b/macosx/JackMachSemaphore.mm#L114) which fails silently if it is hit. Investigation reveals that this is the cause of this failure; each time the client is run, another service port is created, which persist until there are 98, after which `recursiveBootstrapRegister` declines to create any more and semaphore creation fails.

## Proposed Fix
Disclaimer: I am no mach/darwin wizard! However, investigating `bootstrap_register(...)` in [bootstrap.h](https://opensource.apple.com/source/launchd/launchd-328/launchd/src/bootstrap.h.auto.html), the documentation suggests that this approach of registering a service port via the bootstrap server to establish IPC is not recommended (and is actually now deprecated). Our use-case seems to match exactly the described case (3):

```
 * 3) If the code was registering a dynamically generated string and passing
 *    the string to other applications, please rewrite the code to send a Mach
 *    send-right directly.
```

As we are already creating a named shared memory segment to communicate the service name to the client, we could instead simply use the same shared memory segment to pass the send right for the semaphore directly instead, obviating any need to register services with the bootstrap server at all.

I have implemented this change and it seems to resolve the issue; I will create a PR shortly.

p--b:
Previously, JackMachSemaphore would communicate the send right for the
semaphore object from the server to the client via a named service
regstered via `bootstrap_register`. However,

 * This is deprecated (see bootstrap.h) - the recommendation is to pass
 the send right for the port directly, rather than indirecting via the
 service name and the bootstrap server

 * This led to #784, wherein old service names were not cleaned up, and
 the server would abandon attempting to create the 99th service, leading
 to clients failing to connect after they had connected 98 times on a
 given boot of the OS.

 This commit implements the advice from the deprecation notice in
 bootstrap.h, communicating the send right for the semaphore port
 directly via the shared memory segment that we used to use to
 communicate the service name, bypassing the need to register a named
 service entirely. This resolves #784.

**NB**: I am unsure what guarantees JACK makes about the stability of the client/server
interface, or exactly how clients load the dylib containing this functionality - but I assume this
would not be backwards-compatible between different client/server versions (?) - I'm unsure of
what the implications of this are or if some protocol version bump / similar would be required?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -72,6 +72,7 @@ Nedko Arnaudov
 Olaf Hering
 Olivier Humbert
 Paul Davis
+Peter Bridgman
 Peter L Jones
 Pieter Palmers
 Ricardo Crudo
--- a/macosx/JackMachSemaphore.h
+++ b/macosx/JackMachSemaphore.h
@@ -39,12 +39,16 @@ class SERVER_EXPORT JackMachSemaphore : public detail::JackSynchro
     private:
 
         semaphore_t fSemaphore;
-        mach_port_t fBootPort;
 
         int fSharedMem;
-        char* fSharedName;
 
-        bool recursiveBootstrapRegister(int counter);
+        /*! \brief Pointer to shared memory containing semaphore send right port.
+         *
+         * If the semaphore has been Allocate()d (in the case of the server) or Connect()ed (in the
+         * case of the client), and not yet Disconnect()ed or Destroy()ed, a pointer to a shared
+         * memory segment at which a send right for a semaphore can be found. Otherwise, NULL.
+         */
+        mach_port_t* fSharedSemaphoreSend;
 
     protected:
 
@@ -52,7 +56,7 @@ class SERVER_EXPORT JackMachSemaphore : public detail::JackSynchro
 
     public:
 
-        JackMachSemaphore():JackSynchro(), fSemaphore(0), fBootPort(0), fSharedMem(0), fSharedName(NULL)
+        JackMachSemaphore():JackSynchro(), fSemaphore(0), fSharedMem(0), fSharedSemaphoreSend(NULL)
         {}
 
         bool Signal();
--- a/macosx/JackMachSemaphore.mm
+++ b/macosx/JackMachSemaphore.mm
@@ -109,39 +109,6 @@ bool JackMachSemaphore::TimedWait(long usec)
     return (res == KERN_SUCCESS);
 }
 
-bool JackMachSemaphore::recursiveBootstrapRegister(int counter)
-{
-    if (counter == 99)
-        return false;
-
-    kern_return_t res;
-
-    if ((res = bootstrap_register(fBootPort, fSharedName, fSemaphore)) != KERN_SUCCESS) {
-        switch (res) {
-            case BOOTSTRAP_SUCCESS :
-                break;
-
-            case BOOTSTRAP_NOT_PRIVILEGED :
-            case BOOTSTRAP_NAME_IN_USE :
-            case BOOTSTRAP_UNKNOWN_SERVICE :
-            case BOOTSTRAP_SERVICE_ACTIVE :
-                // try again with next suffix 
-                snprintf(fSharedName, sizeof(fName), "%s-%d", fName, ++counter);
-                return recursiveBootstrapRegister(counter);
-                break;
-
-            default :
-                jack_log("bootstrap_register() err = %i:%s", res, bootstrap_strerror(res));
-                break;
-        }
-
-        jack_error("Allocate: can't check in mach semaphore name = %s err = %i:%s", fName, res, bootstrap_strerror(res));
-        return false;
-    }
-
-    return true;
-}
-
 // Server side : publish the semaphore in the global namespace
 bool JackMachSemaphore::Allocate(const char* name, const char* server_name, int value)
 {
@@ -149,38 +116,31 @@ bool JackMachSemaphore::Allocate(const char* name, const char* server_name, int 
     mach_port_t task = mach_task_self();
     kern_return_t res;
 
-    if (fBootPort == 0) {
-        if ((res = task_get_bootstrap_port(task, &fBootPort)) != KERN_SUCCESS) {
-            jack_error("Allocate: Can't find bootstrap mach port err = %s", mach_error_string(res));
-            return false;
-        }
-    }
-
     if ((fSharedMem = shm_open(fName, O_CREAT | O_RDWR, 0777)) < 0) {
-        jack_error("Allocate: can't check in mach shared name = %s err = %s", fName, strerror(errno));
+        jack_error("Allocate: can't open shared memory segment; name = %s err = %s", fName, strerror(errno));
         return false;
     }
 
     struct stat st;
     if (fstat(fSharedMem, &st) != -1 && st.st_size == 0) {
-        if (ftruncate(fSharedMem, SYNC_MAX_NAME_SIZE+1) != 0) {
+        if (ftruncate(fSharedMem, sizeof(mach_port_t)) != 0) {
             jack_error("Allocate: can't set shared memory size in mach shared name = %s err = %s", fName, strerror(errno));
             return false;
         }
     }
 
-    char* const sharedName = (char*)mmap(NULL, SYNC_MAX_NAME_SIZE+1, PROT_READ|PROT_WRITE, MAP_SHARED, fSharedMem, 0);
+    mach_port_t* const sharedSemaphoreSend =
+        (mach_port_t*)mmap(NULL, sizeof(mach_port_t), PROT_READ | PROT_WRITE, MAP_SHARED, fSharedMem, 0);
 
-    if (sharedName == NULL || sharedName == MAP_FAILED) {
-        jack_error("Allocate: can't check in mach shared name = %s err = %s", fName, strerror(errno));
+    if (sharedSemaphoreSend == NULL || sharedSemaphoreSend == MAP_FAILED) {
+        jack_error("Allocate: failed to mmap; name = %s err = %s", fName, strerror(errno));
         close(fSharedMem);
         fSharedMem = -1;
         shm_unlink(fName);
         return false;
     }
 
-    fSharedName = sharedName;
-    strcpy(fSharedName, fName);
+    fSharedSemaphoreSend = sharedSemaphoreSend;
 
     if ((res = semaphore_create(task, &fSemaphore, SYNC_POLICY_FIFO, value)) != KERN_SUCCESS) {
         jack_error("Allocate: can create semaphore err = %i:%s", res, mach_error_string(res));
@@ -188,7 +148,10 @@ bool JackMachSemaphore::Allocate(const char* name, const char* server_name, int 
     }
 
     jack_log("JackMachSemaphore::Allocate name = %s", fName);
-    return recursiveBootstrapRegister(1);
+
+    *fSharedSemaphoreSend = fSemaphore;
+
+    return true;
 }
 
 // Client side : get the published semaphore from server
@@ -198,40 +161,28 @@ bool JackMachSemaphore::ConnectInput(const char* name, const char* server_name)
     kern_return_t res;
 
     // Temporary...
-    if (fSharedName) {
+    if (fSharedSemaphoreSend) {
         jack_log("Already connected name = %s", name);
         return true;
     }
 
-    if (fBootPort == 0) {
-        if ((res = task_get_bootstrap_port(mach_task_self(), &fBootPort)) != KERN_SUCCESS) {
-            jack_error("Connect: can't find bootstrap port err = %s", mach_error_string(res));
-            return false;
-        }
-    }
-
     if ((fSharedMem = shm_open(fName, O_RDWR, 0)) < 0) {
-        jack_error("Connect: can't connect mach shared name = %s err = %s", fName, strerror(errno));
+        jack_error("Connect: can't open shared memory segment; name = %s err = %s", fName, strerror(errno));
         return false;
     }
 
-    char* const sharedName = (char*)mmap(NULL, SYNC_MAX_NAME_SIZE+1, PROT_READ|PROT_WRITE, MAP_SHARED, fSharedMem, 0);
+    mach_port_t* const sharedSemaphoreSend =
+        (mach_port_t*)mmap(NULL, sizeof(mach_port_t), PROT_READ | PROT_WRITE, MAP_SHARED, fSharedMem, 0);
 
-    if (sharedName == NULL || sharedName == MAP_FAILED) {
-        jack_error("Connect: can't connect mach shared name = %s err = %s", fName, strerror(errno));
+    if (sharedSemaphoreSend == NULL || sharedSemaphoreSend == MAP_FAILED) {
+        jack_error("Connect: failed to mmap: name = %s err = %s", fName, strerror(errno));
         close(fSharedMem);
         fSharedMem = -1;
         return false;
     }
 
-    if ((res = bootstrap_look_up(fBootPort, sharedName, &fSemaphore)) != KERN_SUCCESS) {
-        jack_error("Connect: can't find mach semaphore name = %s, sname = %s, err = %s", fName, sharedName, bootstrap_strerror(res));
-        close(fSharedMem);
-        fSharedMem = -1;
-        return false;
-    }
-
-    fSharedName = sharedName;
+    fSharedSemaphoreSend = sharedSemaphoreSend;
+    fSemaphore = *sharedSemaphoreSend;
 
     jack_log("JackMachSemaphore::Connect name = %s ", fName);
     return true;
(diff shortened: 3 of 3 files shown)


falkTX:
Thanks a lot!

Changes seem all good, but we need to confirm that arm64 + x64 interoperability still works.
Do you have an M1 machine?
Since we have automated builds now, we can try those right away. (with the macos-universal artifact)

PS: There are no worries regarding keeping ABI, as JACK lib and server must always be installed together.

p--b:
Unfortunately I don't have access to any M1 hardware at this time - sorry! Hence my slight naïveté on this;

To be clear, is the interoperability question if you have a native aarch64 client talking to a jackd running via rosetta 2?
Or just both client&server running natively? I'm not sure what is supported?

If the latter then `mach_port_t` ought to be fine going via shm; afaict `mach_port_t` is just a pointer to a `struct ipc_port` (https://developer.apple.com/documentation/kernel/mach_port_t, https://opensource.apple.com/source/xnu/xnu-201/osfmk/ipc/ipc_port.h.auto.html) so it's just passing a pointer around.

However I can imagine pointers could look different between the two arches, if that's what's expected to work? I'm afraid I have little idea how rosetta2 actually works, so whether a pointer provided by darwin to an emulated binary could be deref'd by a native binary is... perhaps questionable?! But idk if that's what you mean?

falkTX:
We can have a arm64 jackd server that is able to listen and talk to both native arm64 clients and x64 ones via rosetta2.
I have confirmed this myself.

Though somehow a x64 jackd server cannot establish connections with arm64 clients. I dont know how at this point.

And yeah, this being in shared memory can be questionable for multi-arch access.
But since pointer size of arm64 and x64 are both 8 bytes, maybe it can just work.
We can put the pointer as union so that it always reserves 8 bytes for it, but since Apple is not going back to 32bit, lets not care for now.

I have an M1 myself, will set it up again for use and try this out soon.

p--b:
Brilliant, thanks. Let me know if you need anything else from me.

falkTX:
On some quick testing, this does not seem like it is working at all.
The semaphore waiting fails, which is expected I guess since these semaphores by design are not meant to be used for IPC.
As I understood from the old code, the bootstrap registers a service that allows to use the semaphore globally.

FYI `jack_lsp` is not a good test, since it does not activate the client and thus doesnt need/use the semaphores.
For real testing we need clients that register ports and connect them to something.

I will give this more testing at a later point. But it doesn't sound promising.

p--b:
Interesting. Apologies, I hadn't realised that `jack_lsp` wouldn't exercise this. Indeed trying with `jack_simple_client` it is broken.

I'd be kinda surprised if a semaphore didn't work across processes - however - It seems I'd missed a crucial bit of understanding, which is that each mach task has its own private port table, so a pointer to a given port won't be valid in another task's context unless explicitly moved via a mach message (obligatory [stack overflow clues are available](https://stackoverflow.com/questions/65166614/using-mach-semaphores-as-a-replacement-of-inter-process-anonymous-semaphores)).

From that it seems there's two threads to pull on;

1. continue with the same approach, just send the semaphore port via a `mach_msg` somehow instead of via shm
2. use a SysV named semaphore instead

I can have a look into either/both approaches next week - any thoughts / past experience that would be relevant before I have a go?

falkTX:
not much familiar with point 1, but worth noting perhaps that jack allows more than 1 server to be active at the same time (using different server names). so it needs to remain compatible (or better said, not become incompatible)

there seems to be some issues with named semaphores and sandboxing https://stackoverflow.com/questions/24451774/how-to-enable-ipc-sysv-sem-in-mac-sandboxing
also note that you cant use `sem_timedwait` because it is not implemented on macOS, in case you thought of POSIX ones.

p--b:
Thanks, yes this makes sense. Maybe (1) would be easier, which would involve going back to using the bootstrap server to exchange ports between the processes... digging into it more I have a hunch that what was happening is that as we were registering the _semaphore's_ port with the bootstrap server (not a port directly belonging to the `jackd` _task_), the semaphore's port perhaps continued to exist and never became dead, even when the semaphore was destroyed, which prevented registering a replacement with the bootstrap server (maybe?).

A solution to this might therefore to be to use `mach_port_allocate` to create a port with the receive right belonging to the `jackd` task, and register *that port* with the bootstrap server. The client then looks up that port, sends a message to jackd requesting a send right to the semaphore and providing a send right back to the client, and jackd then finally sends a send right to the semaphore port to the client. A bit convoluted but might work. Major faff would be that the `mach_msg` primitive used to do this is blocking with a timeout, so we'd probably need to spin up a bonus thread in at least the server to handle receiving the request from the client and sending a reply back to it...

The more I think about this the less enamoured with it I become. Maybe I'll go back to trying to work out why the semaphore port wasn't becoming dead when the semaphore was destroyed, because simply replacing the old dead port with a new port with the same service name in the bootstrap server would be much cleaner...

p--b:
Well, I got sniped...

So, after having spent entirely too much time browsing the available source for launchd (which provides the bootstrap server), it seems that launchd performs its own book-keeping of which services are "active", relying on some [slightly hairy logic](https://opensource.apple.com/source/launchd/launchd-842.92.1/src/core.c.auto.html) and notifications from the kernel when things die. I didn't quite get to the bottom of why but it seems that any service which has its port as a semaphore send right will never become "inactive" within launchd - I suspect it's because the semaphore's ports are in a special kernel IPC domain and so they don't appear to belong to the user when launchd is notified that the port has died. As the service never becomes inactive in launchd's eyes, it will never allow the service's ports to be replaced. This means that any approach that relies on registering a semaphore send right with the bootstrap server directly won't work (at least until this ?bug? is fixed in macOS).

Investigating non-semaphore ports being registered with the bootstrap server showed them being cleaned up correctly when jackd died, which led me back to my previous thought of registering normally allocated ports and doing a `mach_msg` handshake. I've done a quick and dirty test implementation and it seems to work (testing with `jack_simple_client` this time ;) ). It is now very much bedtime but I will tidy it up and update this PR sometime ~tomorrow~ later today.

tldr; fixed via new approach based on `mach_msg`, updated PR incoming.

PR #788, opened 2021-08-08
macOS: Pass JackMachSemaphore send right via mach_msg IPC

### LINKED ISSUE #784: macOS: clients can only connect 98 times per boot

p--b:
Previously, JackMachSemaphore would communicate the send right for the
semaphore object from the server to a client via a named service
registered via `bootstrap_register`. However, to do this, it would
register the semaphore's port as the service port directly.

In theory this ought to be fine, however in practice, macOS `launchd`,
which provides the `bootstrap_register` interface, does not correctly
detect when such a port becomes dead, and incorrectly believes that the
service that it provides is forever alive, even past the end of the
`jackd` process' (and therefore the semaphore's) existence. This seems
to be *specific* to semaphore ports, as `launchd` is expecting a
standard IPC port, owned by the task, not the kernel. This prevents
`jackd` from later registering another service with the same name, as
launchd rejects the registration as conflicting with an active service.

To get around this, `jackd` previously added a counter to the end of the
named service registrations, allowing old services to remain in the
system until the end of the session. To prevent things getting out of
hand, this was capped at 98 service registrations for a given semaphore
name. This led to #784, in which running a client for the 99th time
resulted in the semaphore creation failing and the client failing to
connect.

As launchd outlives multiple runs of `jackd`, this situation persisted
across restarts of jackd, requiring a restart of the user's session
(i.e. a reboot) to fix.

An initial attempt at fixing this (see #785) tried passing the port
rights directly via shared memory, however mach is too clever for us and
foils that plan by having port names be looked up in a per-task table
(sensible when you think about it).

In this commit, we use mach IPC messages to transfer the send right for
the semaphore from the server to the client. By registering a standard IPC
port with the bootstrap server, the service registrations are correctly torn
down when the ports are destroyed.

It works something like this:

* Server creates IPC port and registers it globally via `bootstrap_register`
* Server listens on IPC port for messages
* Client looks up IPC port via `bootstrap_look_up`
* Client sends it a message
* Server replies with a message containing a send right to the
semaphore's port
* Client is then free to use the semaphore port as before.

This resolves #784.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -72,6 +72,7 @@ Nedko Arnaudov
 Olaf Hering
 Olivier Humbert
 Paul Davis
+Peter Bridgman
 Peter L Jones
 Pieter Palmers
 Ricardo Crudo
--- a/common/wscript
+++ b/common/wscript
@@ -117,6 +117,7 @@ def build(bld):
             '../posix/JackPosixMutex.cpp',
             '../macosx/JackMachThread.mm',
             '../macosx/JackMachSemaphore.mm',
+            '../macosx/JackMachSemaphoreServer.mm',
             '../posix/JackSocket.cpp',
             '../macosx/JackMachTime.c',
             ]
--- a/macosx/JackMachSemaphore.h
+++ b/macosx/JackMachSemaphore.h
@@ -22,6 +22,10 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
 #include "JackCompilerDeps.h"
 #include "JackSynchro.h"
+
+#include "JackMachThread.h"
+#include "JackMachSemaphoreServer.h"
+
 #include <mach/mach.h>
 #include <servers/bootstrap.h>
 #include <mach/semaphore.h>
@@ -38,13 +42,24 @@ class SERVER_EXPORT JackMachSemaphore : public detail::JackSynchro
 
     private:
 
+        /*! \brief A mach send right to the mach semaphore, or MACH_PORT_NULL if not yet Allocate()d
+         * (server) or Connect()ed (client). */
         semaphore_t fSemaphore;
+
+        /*! \brief The bootstrap port for this task, or MACH_PORT_NULL if not yet obtained. */
         mach_port_t fBootPort;
 
-        int fSharedMem;
-        char* fSharedName;
+        /*! \brief The IPC port used to pass the semaphore port from the server to the client, and
+         * for the client to request that this occurs. MACH_PORT_NULL if not yet created (server) or
+         * looked up (client). */
+        mach_port_t fServicePort;
 
-        bool recursiveBootstrapRegister(int counter);
+        /*! \brief On the server, if allocated, a runnable semaphore server which listens for IPC
+         * messages and replies with a send right for a semaphore port. */
+        JackMachSemaphoreServer* fSemServer;
+
+        /*! \brief On the server, if allocated, a thread that runs \ref fSemServer. */
+        JackMachThread* fThreadSemServer;
 
     protected:
 
@@ -52,7 +67,13 @@ class SERVER_EXPORT JackMachSemaphore : public detail::JackSynchro
 
     public:
 
-        JackMachSemaphore():JackSynchro(), fSemaphore(0), fBootPort(0), fSharedMem(0), fSharedName(NULL)
+        JackMachSemaphore():
+            JackSynchro(),
+            fSemaphore(MACH_PORT_NULL),
+            fBootPort(MACH_PORT_NULL),
+            fServicePort(MACH_PORT_NULL),
+            fSemServer(NULL),
+            fThreadSemServer(NULL)
         {}
 
         bool Signal();
--- a/macosx/JackMachSemaphore.mm
+++ b/macosx/JackMachSemaphore.mm
@@ -18,12 +18,18 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
 #include "JackMachSemaphore.h"
+#include "JackMachUtils.h"
 #include "JackConstants.h"
 #include "JackTools.h"
 #include "JackError.h"
-#include <fcntl.h>
-#include <stdio.h>
-#include <sys/mman.h>
+
+#include <mach/message.h>
+
+#define jack_mach_error(kern_result, message) \
+        jack_mach_error_uncurried("JackMachSemaphore", kern_result, message)
+
+#define jack_mach_bootstrap_err(kern_result, message, name) \
+        jack_mach_bootstrap_err_uncurried("JackMachSemaphore", kern_result, message, name)
 
 namespace Jack
 {
@@ -42,7 +48,7 @@ void JackMachSemaphore::BuildName(const char* client_name, const char* server_na
 
 bool JackMachSemaphore::Signal()
 {
-    if (!fSemaphore) {
+    if (fSemaphore == MACH_PORT_NULL) {
         jack_error("JackMachSemaphore::Signal name = %s already deallocated!!", fName);
         return false;
     }
@@ -60,7 +66,7 @@ bool JackMachSemaphore::Signal()
 
 bool JackMachSemaphore::SignalAll()
 {
-    if (!fSemaphore) {
+    if (fSemaphore == MACH_PORT_NULL) {
         jack_error("JackMachSemaphore::SignalAll name = %s already deallocated!!", fName);
         return false;
     }
@@ -79,7 +85,7 @@ bool JackMachSemaphore::SignalAll()
 
 bool JackMachSemaphore::Wait()
 {
-    if (!fSemaphore) {
+    if (fSemaphore == MACH_PORT_NULL) {
         jack_error("JackMachSemaphore::Wait name = %s already deallocated!!", fName);
         return false;
     }
@@ -93,7 +99,7 @@ bool JackMachSemaphore::Wait()
 
 bool JackMachSemaphore::TimedWait(long usec)
 {
-    if (!fSemaphore) {
+    if (fSemaphore == MACH_PORT_NULL) {
         jack_error("JackMachSemaphore::TimedWait name = %s already deallocated!!", fName);
         return false;
     }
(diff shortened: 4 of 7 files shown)


p--b:
@falkTX as promised - PTAL

falkTX:
Overall looks good, thanks for the extra code comments!
One thing though, there is no cleanup on error at all in the JackMachSemaphore class.
If bootstrap process goes ok but thread fails to start, we need to cleanup the bootstrap stuff. Otherwise after single 1 error in the system jack becomes unusable for certain clients.

p--b:
Ah, so if `Allocate()` returns `false` then `Destroy()` won't be called? I see what you mean - will add

falkTX:
> Ah, so if `Allocate()` returns `false` then `Destroy()` won't be called?

yes. like you dont call free() on a failed malloc.
thanks

p--b:
This should be better, PTAL

p--b:
Looks like that CI failure might be spurious?

falkTX:
yes, for some reason xvfb-run randomly fails to start sometimes

sletz:
Nice to see this old Mach based code properly reworked. @p--b thanks for that, since having a precise knowledge of those exotic Mach APIs is quite rare ((-;

p--b:
@sletz I wouldn't assume I have too much precise knowledge... I just really enjoy reading old Apple opensource listings ;)

falkTX:
Gave this a try, all seems to be working as far as I can tell.
Merging, thanks a lot!

vrslev:
@p--b @falkTX  When this fix will be included in stable release?

falkTX:
in 3 days

vrslev:
@falkTX Thanks!

PR #796, opened 2021-09-27
using wrapper function

ihsinme:
call, with validation.


falkTX:
Just verified the automated builds keep working, seems good, thanks!

Issue #459, opened 2019-05-09
S24_LE not working for alsa driver


mr-berndt:
I understand that from jackd 1.9.11 on jack should be able to negotiate S24_LE with the card, if S32_LE is unavailable.

Now when I use a HAT on a Raspberry Pi which outputs SPDIF and can do S24_LE, like the Allo Digi One, this seems to fail:

Making sure the card can do it:
`# cat /proc/asound/card0/pcm0p/sub0/hw_params`
```
access: MMAP_INTERLEAVED
format: S24_LE
subformat: STD
channels: 2
rate: 192000 (192000/1)
period_size: 6960
buffer_size: 27840
```

Audio plays fine.

When starting jack with `jackd -R -P 50 -d alsa -P -r 96000 -p 8192 -n 3` I get:
```
creating alsa driver ... hw:0|-|8192|3|96000|0|0|nomon|swmeter|-|32bit
configuring for 96000Hz, period = 8192 frames (85.3 ms), buffer = 3 periods
ALSA: final selected sample format for playback: 24bit little-endian
```
and terribly noisy output (sounds like radio inbetween stations). When lowering the volume the sound becomes clear.

When starting jackd with `jackd -R -P 50 -d alsa -S -P -r 96000 -p 8192 -n 3`
I get
```
creating alsa driver ... hw:0|-|8192|3|96000|0|0|nomon|swmeter|-|16bit
configuring for 96000Hz, period = 8192 frames (85.3 ms), buffer = 3 periods
ALSA: final selected sample format for playback: 16bit little-endian
```
and the sound is fine but downconverted to 16 bits.

Is there some switch for S24_LE I am missing or is this a bug?

7890:
Please try the most current git (master branch) and check if this is still an issue.
There was a related bug in the repository for a short period of time which is now fixed again. If you test it with current git, we can rule out any side-effects from that.

mr-berndt:
I used the latest commit from Apr. 27 - f7f2244b07ee0a723853e838de85e25471b8903f

7890:
ok it's unrelated to the previous bug in that case.

mr-berndt:
Did you have a chance yet to look into it?

7890:
What do you get with a simple `jackd --verbose -dalsa` ?
When jackd is running, `cat /proc/asound/card0/pcm0c/sub0/hw_params`

romanticamaj:
[...]

PR #798, opened 2021-10-01
ALSA: convert properly S24_LE

camille-sound4:
ALSA S24_LE is using the **lower** 3 bytes, not the upper.

> [Signed 24 bit Little Endian using low three bytes in 32-bit word](https://git.alsa-project.org/?p=alsa-lib.git;a=blob;f=include/pcm.h#l140)

This patch adds the memops for all 32l24 conversions like 32u24, and use them for ALSA driver only.
Maybe some other driver should use this too (OSS/Boomer) but I can not test them.
SIMD optimizations have been checked on ARM NEON and x86 SSE2.

Fixes jackaudio/jack2#459

### DIFF
--- a/common/memops.c
+++ b/common/memops.c
@@ -137,6 +137,15 @@
 		(d) = f_round ((s) * SAMPLE_24BIT_SCALING) << 8;\
 	}
 
+#define float_24l32(s, d)                                              \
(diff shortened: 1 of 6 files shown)


falkTX:
Verified this to work, thanks a lot!
This time it doesn't seem to break anything else from what I can tell.

PR #800, opened 2021-10-22
FreeBSD support, specific OSS driver backend 

0EVSG:
Revive the efforts for FreeBSD support, based on #400 with the obstacles (hopefully) fixed:

 - Symlinks replaced by relay include header files.
 - Indents compliant with jack2 style, tabs removed.

Also contains a big overhaul of the OSS driver backend for FreeBSD:

- Separate JackOSSDriver.* files because of extensive changes specific to FreeBSD.
- New `poll()` based wait mechanism, closer to the ALSA backend implementation.
- Allows jack2 to operate in "asynchronous" mode.
- Does OSS buffer management to get stable latency, despite drift or XRUNs.
- Use 24bit packed sample conversion as required on FreeBSD.
- Gracefully handle cases where requested sample formats are not available.

These changes are already in FreeBSD ports since august and user feedback has been positive. In my own experience, the stable latency makes a great difference in usability, e.g. for recording multiple takes.

The FreeBSD [port](https://www.freshports.org/audio/jack/) currently fetches sources from my repo, I'd love to get this into the main repo.


mekanix:
Hello!

I would like to responsibility of maintaining FreeBSD support. As @0EVSG wrote most of the patch, if he would like to be that person, that would be great, otherwise, I'd just like to take some of the burden off his back.

0EVSG:
Hi @mekanix, thanks for your offer - while I don't really mind, it's probably easier if I take the responsibility of maintaining FreeBSD support. I'm currently working on making real time priority accessible to users in FreeBSD. If that succeeds, I'll have to make some further changes regarding Jack threads anyway.

As suggested I'll join you and @falkTX on IRC to discuss the details, probably tomorrow.

0EVSG:
@falkTX: I'd like to merge FreeBSD support next week, to give us some time before the release. Maybe you or someone else from the Team (who?) wants to have a look at it first?

falkTX (approved):

falkTX:
Just had a quick look at it, should be quite fine to merge at any time, thanks for the efforts.
Only a minor commenting out is a bit odd, but the code was defining _POSIX_SOURCE which is not ideal anyway (nowadays the compiler does that for us afaik)

0EVSG:
> Just had a quick look at it, should be quite fine to merge at any time, thanks for the efforts. Only a minor commenting out is a bit odd, but the code was defining _POSIX_SOURCE which is not ideal anyway (nowadays the compiler does that for us afaik)

Thanks for the hint, _POSIX_SOURCE was commented out by @mekanix, I'll have a look at it before I merge.

0EVSG:
Added commit abe77eaba4285622bcad3ddf8f1fb7be2a6c6c58 with an explanation for the removal of _POSIX_C_SOURCE. This line was introduced by #582, and then commented out by @mekanix to make it compile on FreeBSD.

The _POSIX_C_SOURCE macro is meant to restrict the C library to a certain POSIX standard, not to enable features. This is inappropriate here because:

1. Jack already uses non-POSIX functions. E.g. JackPosixSemaphore.cpp references gettimeofday(), which is part of the X/Open specs.

2. C++ does not define a POSIX level to be based on. Headers from the C++ standard library may (and do) break when the POSIX feature level is restricted.

The canonical way is to only set the feature level through the "--std=gnu++11" compile options.

The commit itself is a no-op, not sure why the Windows CI builds fail - the error messages seem unrelated.

PR #803, opened 2021-10-26
Fix grouping for latest doxygen.

traud:
Since doxygen 1.8.16, opening and closing a group must not be done as C comment but as doxygen command. In other words, not one but two asterisk characters are required so that doxygen finds a group. This fixed:
```
common/jack/control.h:659: warning: end of file while inside a group
common/jack/jack.h:1478: warning: end of file while inside a group
common/jack/midiport.h:198: warning: end of file while inside a group
common/jack/thread.h:161: warning: end of file while inside a group
common/jack/transport.h:248: warning: end of file while inside a group
common/jack/weakjack.h:126: warning: end of file while inside a group
```
Then, the removal of a superfluous closing at jack.h:560 fixed:
```
common/jack/jack.h:622: warning: unbalanced grouping commands
```
The changes at line 1054 and 1417 are just cosmetic, to make the declaration of [doxygen grouping](https://www.doxygen.nl/manual/grouping.html) consistent, use one style throughout the jack2 headers.

Tested via:
```
$ ./waf configure --doxygen=yes
$ ./waf clean
$ ./waf
```

### DIFF
--- a/common/jack/control.h
+++ b/common/jack/control.h
@@ -646,7 +646,7 @@ jack_log(
 	const char *format,
 	...);
 
-/* @} */
+/**@}*/
 
 #if 0
 { /* Adjust editor indent */
--- a/common/jack/jack.h
+++ b/common/jack/jack.h
@@ -230,7 +230,7 @@ int jack_get_client_pid (const char *name) JACK_OPTIONAL_WEAK_EXPORT;
  */
 jack_native_thread_t jack_client_thread_id (jack_client_t *client) JACK_OPTIONAL_WEAK_EXPORT;
 
-/*@}*/
+/**@}*/
 
 /**
  * @param client pointer to JACK client structure.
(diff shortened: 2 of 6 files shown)


traud:
Anything blocking this? Do I have to find a maintainer myself via some mailing list, IRC, or ‥?

falkTX:
I am just busy overall, have to see which things require immediate attention and which ones can be left for later.

Did you check the jack1 headers to see how it is done in there?
Eventually jack1 and jack2 headers should be merged, so would be great to have this in common between them.
jack1 headers are in https://github.com/jackaudio/headers/

traud:
Yes, I checked those, too. I thought you backport that change then, especially because I do not use the old headers personally. I love to fix issues in things I benefit from, but I cannot rescue the (whole) world.

falkTX:
It is just another PR in another project, not the whole world..

Changes to core files need to be done carefully now, otherwise projects like https://github.com/jackaudio/jack-example-tools will never be finalized.
Having the same files in both jack1 and jack2 is a necessity going forward.

traud:
You can hijack the change and create a Pull Request there. I do not know how to couple two Pull Requests over two different repositories.

falkTX:
We can see about those later on.
But yeah, idea is to eventually merge jack1 and jack2 headers into a single repo.

PR #811, opened 2021-11-12
Add 32 bit support for ALSA driver

### LINKED ISSUE #810: Jack does not support 32 bit mode in ALSA
Jack always assumes only 24 out of 32 bits are used in its ALSA driver, even if the sample format is 32_LE.

Previously this only caused a dynamic range reduction, but with the latest fixes to 24 bit handling in alsa (see #459 and #798 ), 32 bit cards suffer a gain reduction of 48dB in their outputs and clip very early in their inputs.

Reproduced both on Aarch64 and x86_64, both with linux, using builds from commit 0727bbc386288f6ff5cb6876d2321836d562221e
used a behringer UMC404HD sound card to test on x86_64

claudiocabral:
Fixes #810 

adds new memops functions to convert 32 ints to float and float to 32 bit ints
uses doubles as intermediaries for a sign extension.

I haven't added SIMD instructions for x86 as I'm not too familiar with them and was having issues writing something that works.

Neon instructions were close enough to the original 24 bit versions for me to be able to port them over.

### DIFF
--- a/common/memops.c
+++ b/common/memops.c
@@ -84,8 +84,8 @@
 
 #define SAMPLE_32BIT_MAX   2147483647
 #define SAMPLE_32BIT_MIN   -2147483647
-#define SAMPLE_32BIT_MAX_F  2147483647.0
-#define SAMPLE_32BIT_MIN_F  -2147483647.0
+#define SAMPLE_32BIT_MAX_D  2147483647.0
+#define SAMPLE_32BIT_MIN_D  -2147483647.0
 
 #define SAMPLE_24BIT_MAX  8388607  
 #define SAMPLE_24BIT_MIN  -8388607 
@@ -153,15 +153,14 @@
 		(d) = f_round ((s) * SAMPLE_24BIT_SCALING);                    \
 	}
 
-#define float_32(s, d)                                                 \
-	if ((s) <= NORMALIZED_FLOAT_MIN) {\
-		(d) = SAMPLE_32BIT_MIN;                                        \
-	} else if ((s) >= NORMALIZED_FLOAT_MAX) {\
-		(d) = SAMPLE_32BIT_MAX;                                        \
-	} else {\
-        double extended_value = ((double)s) * SAMPLE_32BIT_SCALING;      \
-		(d) = d_round (extended_value);                                \
-	}
+#define float_32(s, d)												\
+	do {															\
+		double clipped = fmin(NORMALIZED_FLOAT_MAX,					\
+				fmax((double)(s), NORMALIZED_FLOAT_MIN));			\
+		double scaled = clipped * SAMPLE_32BIT_MAX_D;				\
+		(d) = d_round(scaled);										\
+	}																\
+	while (0)
 
 /* call this when "s" has already been scaled (e.g. when dithering)
  */
@@ -296,14 +295,14 @@ void sample_move_dS_floatLE (char *dst, jack_default_audio_sample_t *src, unsign
    
    S      - sample is a jack_default_audio_sample_t, currently (October 2008) a 32 bit floating point value
    Ss     - like S but reverse endian from the host CPU
-   32     - sample is an signed 32 bit integer value
-   32u24  - sample is an signed 32 bit integer value, but data is in upper 24 bits only
+   32     - sample is a signed 32 bit integer value
+   32u24  - sample is a signed 32 bit integer value, but data is in upper 24 bits only
    32u24s - like 32u24 but reverse endian from the host CPU
-   32l24  - sample is an signed 32 bit integer value, but data is in lower 24 bits only
+   32l24  - sample is a signed 32 bit integer value, but data is in lower 24 bits only
    32l24s - like 32l24 but reverse endian from the host CPU
-   24     - sample is an signed 24 bit integer value
+   24     - sample is a signed 24 bit integer value
    24s    - like 24 but reverse endian from the host CPU
-   16     - sample is an signed 16 bit integer value
+   16     - sample is a signed 16 bit integer value
    16s    - like 16 but reverse endian from the host CPU
 
    For obvious reasons, the reverse endian versions only show as source types.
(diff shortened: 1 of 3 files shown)


falkTX:
The last time this was attempted there were causes of bad audio and we had to revert.
so we need to go about this very carefully.

claudiocabral:
agreed, but as it is now 32 bit audio is essentially broken.
On my local tests audio seems good

claudiocabral:
after looking into the generated code using compiler explorer, it seems that
my hand rolled NEON is worst then using the compiler flag `-ftree-loop-vectorize`.
It also breaks compilation for 32 bit architectures.
https://godbolt.org/z/dYb5KsojM

I'll remove it, and if necessary someone with more expertise in SIMD then I have can make a PR to include it in the future

claudiocabral:
I've also looked into the previous attempt to add 32 bit support.
They tried to use 32 bit floats with some clever bitshifting to sign extend and accomodate for the 32 bit integers,
but 32 bit floats can't represent integers that big. That was probably the cause of bad audio in the pass.

I've decided to go for the big hammer and temporarily conver to doubles, which can accomodate for the extra bits.

Please let me know if there's any sort of testing that I can do/add to increase the security of merging this patch.

common/memops.c
```diff
@@ -81,6 +82,11 @@
    advice from Fons Adriaensen: make the limits symmetrical
  */
 
+#define SAMPLE_32BIT_MAX   2147483647
+#define SAMPLE_32BIT_MIN   -2147483647
+#define SAMPLE_32BIT_MAX_F  2147483647.0
```
imaami, line 87: These should probably be suffixed with `_D` instead of `_F` because the existing macro names obviously refer to them being float literals.

common/memops.c
```diff
@@ -146,6 +153,16 @@
 		(d) = f_round ((s) * SAMPLE_24BIT_SCALING);                    \
 	}
 
+#define float_32(s, d)                                                 \
```
imaami, line 156: This whole if-else block needs to be wrapped inside `do { ... } while(0)` to ensure that using it behaves like a function call would.

common/memops.c
```diff
@@ -146,6 +153,16 @@
...
+	if ((s) <= NORMALIZED_FLOAT_MIN) {\
+		(d) = SAMPLE_32BIT_MIN;                                        \
+	} else if ((s) >= NORMALIZED_FLOAT_MAX) {\
+		(d) = SAMPLE_32BIT_MAX;                                        \
+	} else {\
+        double extended_value = ((double)s) * SAMPLE_32BIT_SCALING;      \
```
imaami, line 162: The temporary variable is unnecessary. The same goes for parentheses around `(double)s` because a cast takes precedence over multiplication. The macro parameter should be parenthesized on the other hand, so `(double)(s)`.

common/memops.c
```diff
@@ -274,6 +316,7 @@ void sample_move_dS_floatLE (char *dst, jack_default_audio_sample_t *src, unsign
    
    S      - sample is a jack_default_audio_sample_t, currently (October 2008) a 32 bit floating point value
    Ss     - like S but reverse endian from the host CPU
+   32     - sample is an signed 32 bit integer value
```
imaami, line 299: s/an signed/a signed/

common/memops.c
claudiocabral: agreed, I kept it like this for consitency with the other macros.
Personally I'd forgo the macro all together

common/memops.c
claudiocabral: good point!

common/memops.c
claudiocabral: will fix

common/memops.c
claudiocabral: I like the temporary variable for clarity; any compiler should optimize that away.
also like the extra parenthesis on the cast for clairty;

that said, will gladly do the changes if necessary
(and will include the parenthesize the `s` tomorrow)

claudiocabral: @imaami ended up rewritting quite a bit of the macro, would be thankful it if you'd review it again :)

imaami: Hi! I happened to be looking at my phone and saw a notification, this is not my usual reaction time. :)

In other words, I'll try to keep this in mind and have a look when I can; right now I have work stuff to do.

As a side note, I noticed there's quite a lot of casting from `char *` to `int32_t *` and back. I find it unsettling as it's a potential cause of alignment bugs. And I don't mean your PR specifically, the whole file just seems a little sus.

claudiocabral: sounds good, just don't want to lose traction here :)
 I don't know how much we can do about the casts... the ALSA driver in particular forces you to cast back and forth between pointer types

imaami: Do you happen to know what guarantees, if any, ALSA gives wrt. the buffers' alignment?

claudiocabral: yes, alsa gives us:
a pointer to the beginning of the buffer
the number of bits between each word
the size of each word and which bits to read (in case of 24 bit words packed in 32 bits)

takaswie:
In regard of ALSA PCM interface, I add some comment about the patch in #810 . I'd like developers to read it for the case to handle `SNDRV_PCM_FORMAT_[S|U]32_[L|B]E` adequately.

falkTX:
Would be great to merge this, but it is a little bit out of my expertise to be honest.
Is this good to merge as-is, or is there more work needed?

claudiocabral:
I have a couple of things @takaswie  asked me to go through (will do it after the holidays)
and the windows builds have failed (seems unrelated to this, but can you confirm @falkTX ?)

falkTX:
> windows builds have failed (seems unrelated to this, but can you confirm @falkTX ?)

yes, unrelated.
github updated the CI VM images and broke quite a few things with it..

imaami:
I'll try to look at this during the holidays, this is on a best-effort basis though because I might have things to do that take precedence. Or I may forget. Still very much with you guys in spirit. :)

falkTX:
@claudiocabral please rebase against latest develop, so we have CI working for these changes.
I am about to do a release though, wondering if pushing these changes there or not...

claudiocabral:
@falkTX that's a good question.
Without these changes 32bit devices are broken, with them there are untested devices (though it should all work)
We're also still missing @takaswie review.
How bad is it if we need to undo or patch them after the release?

falkTX:
Undoing is quite fine, we did it before already.
We can put this in a release as a way to get more widespread testing, otherwise we are waiting forever.

The CI stuff passes, so I will just go with it.
Thanks!

If anyone spots anything that is broken by accident, please open a ticket.

claudiocabral:
happy to contribute :)

PR #813, opened 2021-11-27
OpenSL ES driver. Changing the queue processing strategy.

PavelAtr:
Queue handling has been moved to Callback functions. The write / read functions only fill the audio data buffers. This allows for continuous flow of data to / from queues.
That provides continuous sound without clicks and distortion. The driver was tested on Android phone Xiaomi Redmi 8 Note with DAW Ardor and showed good sound quality.

### DIFF
--- a/android/opensl_io.c
+++ b/android/opensl_io.c
@@ -28,6 +28,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 #include "opensl_io.h"
+#include <string.h>
+#include <unistd.h>
 //#define CONV16BIT 32768
 //#define CONVMYFLT (1./32768.)
 #define CONV16BIT 32640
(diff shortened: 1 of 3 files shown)


android/opensl_io.c
```diff
@@ -414,39 +435,66 @@ double android_GetTimestamp(OPENSL_STREAM *p){
 void bqRecorderCallback(SLAndroidSimpleBufferQueueItf bq, void *context)
 {
   OPENSL_STREAM *p = (OPENSL_STREAM *) context;
+
+  while ((p->inputBufferSize[0] > 0) && (p->inputBufferSize[1] > 0))
+    usleep(1000);
```
falkTX, line 440: is this really the best way? couldnt a semaphore or similar be used to keep things in sync?
sleeping until the buffers are empty does not seem like a good idea...

PavelAtr: To be honest, I don't know if it could be done better. I was not satisfied with the initial sound quality of the driver, and my version is as follows. In any case, in the examples I've seen, the callback function itself replenishes the queue, and not the handler from the outside.

PavelAtr: usleep (1000) only happens when there is no sound playing, it does not happen during playback. The examples I've seen use silence buffer playback instead. What is more beautiful I do not know.

PavelAtr: In the absence of sound, usleep allows you to reduce the load on the cpu up to 1-2%, if you do not do this, then a fast cycle will load the cpu.

Issue #757, opened 2021-05-03
Metadata database does not honour the user set by `JACK_PROMISCUOUS_SERVER`


gnif:
When using Jack2 in a headless configuration with `JACK_PROMISCUOUS_SERVER` the Berkely database is created with the default permissions and ownership of the process, in my instance `drwxr-x---` `root:root`. This breaks the metadata feature for any clients running as promiscuous user.

## Environment

* **JACK Version:** 1.9.17
* **Operating System:** Debian
* **Installation:** Compiled from Debian source package with `--db` and `libdb-dev` installed.

## Steps To Reproduce

```bash
# export JACK_NO_AUDIO_RESERVATION=1
# export JACK_PROMISCUOUS_SERVER=someuser
# export $(dbus-launch)
# exec /usr/bin/jackd -R -P84 -dalsa -s -dhw:0 -r48000 -p512 -n2 -H -M
```

### Expected vs. actual behavior

`/dev/shm/jack_db` and it's contents should be owned by the group specified by `JACK_PROMISCUOUS_SERVER` and be writeable by the group, instead it is owned by the user jack was run as (root:root) and is not group writeable.

catniptwinz:
I'm seeing a similar issue with `JACK_PROMISCUOUS_SERVER` in which `/dev/shm/jack_db` is owned by the expected user/group but is not group writeable.

Arch ARM, jack2 1.9.17 installed with pacman, `JACK_PROMISCUOUS_SERVER=audio`. I'm starting jackd with a systemd service, running as a dedicated system user.

Starting a client (tested with supernova) as another user in the audio group throws `cannot open DB environment: Permission denied`, which I'd expect given that `ls -lah /dev/shm/jack_db` shows:

```
drwxr-x--- 2 jack audio  120 Jun 24 02:09 .
drwxrwxrwt 3 root root   180 Jun 24 02:09 ..
-rw-r----- 1 jack audio 296K Jun 24 02:09 __db.001
-rw-r----- 1 jack audio  40K Jun 24 02:09 __db.002
-rw-r----- 1 jack audio  48K Jun 24 02:09 __db.003
-rw-r--r-- 1 jack audio  12K Jun 24 02:09 metadata.db
```

If I manually add group write permissions after starting jackd (`chmod g+w /dev/shm/jack_db`), I'm able to start a client with no error message.

Is this expected behavior? If so what can I do to ensure that the database permissions are set correctly?

gnif:
[...]

PR #818, opened 2021-12-05
Override permission issues by separating users metadata

### LINKED ISSUE #569: Client unable to connect to JACK server v.1.9.14
Is there a notable change in JACK's API between version `1.9.12` and `1.9.14`?
Clients that were able to connect to `1.9.12` are now unable to do so with `1.9.14`. Example with Rivendell ElvishArtisan/rivendell#569: reverting to `1.9.12` makes it able to connect again, even with a full recompile against the correct headers.
On my Fedora 32 machine with my downgraded JACK version `1.9.12`, now PulseAudio plugins can't connect anymore. I guess it has something to do with a major change between versions.
Any hint?

trap000d:
I have issues on computer wits several user accounts and jack2 and jack2-dbus installed. It looks like dbus uses UID and GID of current user and mask 027 for starting jack2
```
[root@arch shm]# ls -la
total 72
drwxr-x---  2 me me   120 Dec  5 11:57 jack_db

[root@arch shm]# ls -la jack_db/
total 324
drwxr-x--- 2 me me    120 Dec  5 11:57 .
drwxrwxrwt 4 root     root        100 Dec  5 12:58 ..
-rw-r----- 1 me me  393216 Dec  5 11:57 __db.001
-rw-r----- 1 me me  49152 Dec  5 11:57 __db.002
-rw-r----- 1 me me  49152 Dec  5 11:57 __db.003
-rw-r--r-- 1 me me  12288 Dec  5 11:57 metadata.db

```
Effectively, other user can't connect to metadata. There is a lot of reports out of github in regarding this issue as well.
Proposed PR is about to handle metadata of all users separately, as it was in jack1 (and is implemented in other software). Hopefully it should fix #569 and #757.

### DIFF
--- a/common/JackMetadata.cpp
+++ b/common/JackMetadata.cpp
@@ -21,6 +21,7 @@
 #include "JackMetadata.h"
 
 #include "JackClient.h"
+#include "JackTools.h"
 
 #include <string.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


falkTX:
Makes sense, thanks!

PR #825, opened 2022-01-06
Optional example clients and tools

### LINKED ISSUE #805: Devendor all example-clients and tools
## Describe the bug

The [jack-example-tools](https://github.com/jackaudio/jack-example-tools) will soon replace the use and need of [example-clients](https://github.com/jackaudio/example-clients) and [tools](https://github.com/jackaudio/tools) in the context of the jack1 project.
For the new project to be able to also fully replace the [example-clients](https://github.com/jackaudio/jack2/tree/0727bbc386288f6ff5cb6876d2321836d562221e/example-clients) and [tools](https://github.com/jackaudio/jack2/tree/0727bbc386288f6ff5cb6876d2321836d562221e/tools) of the jack2 project, they need to be consolidated with the new code base and afterwards fully devendored from this project.

This way we will be able to provide them for any implementation of jack (also pipewire, see relevant upstream ticket: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1666) in the future and at the same time enable for an easier development of the jack2 code.

## Environment

* **JACK Version:** 1.9.19
* **Operating System:** any
* **Installation:** n/a

## Steps To Reproduce

* compare jack1 and jack2 installations: the example-clients and tools differ (slightly)
* install pipewire-jack: the example-clients and tools are missing and can not be installed, because when pipewire is built as a full jack replacement, it conflicts with jack1 and jack2 (depending on package splitting)

The jack-example-tools can be installed while switching jack implementations (jack1, jack2 or pipewire-jack).
Currently only different versions of the example-clients and tools can be installed alongside jack1 and jack2, but are missing when using pipewire-jack.

dvzrv:
This allows for omitting the build and installation of libraries, executables and man pages provided by [jack-example-tools](https://github.com/jackaudio/jack-example-tools), by setting the new configure flag `--example-tools` (which currently defaults to `"yes"`) to `"no"`.

Additionally, this PR fixes many syntax problems in the top-level wscript file.

Fixes #805

### DIFF
--- a/wscript
+++ b/wscript
@@ -134,6 +134,12 @@ def options(opt):
             'celt',
             help='Build with CELT')
     celt.add_function(check_for_celt)
+    opt.add_auto_option(
+            'example-tools',
+            help='Build with jack-example-tools',
+            conf_dest='BUILD_JACK_EXAMPLE_TOOLS',
+            default=True,
+    )
 
     # Suffix _PKG to not collide with HAVE_OPUS defined by the option.
     opus = opt.add_auto_option(
@@ -296,9 +302,9 @@ def configure(conf):
         else:
             conf.env['SYSTEMD_USER_UNIT_DIR'] = None
 
-
-    conf.recurse('example-clients')
-    conf.recurse('tools')
+    if conf.env['HAVE_JACK_EXAMPLE_TOOLS']:
+        conf.recurse('example-clients')
+        conf.recurse('tools')
 
     # test for the availability of ucontext, and how it should be used
     for t in ['gp_regs', 'uc_regs', 'mc_gregs', 'gregs']:
@@ -839,13 +845,14 @@ def build(bld):
 
     build_drivers(bld)
 
-    bld.recurse('example-clients')
-    bld.recurse('tools')
+    if bld.env['HAVE_JACK_EXAMPLE_TOOLS']:
+        bld.recurse('example-clients')
+        bld.recurse('tools')
 
     if bld.env['IS_LINUX'] or bld.env['IS_FREEBSD']:
         bld.recurse('man')
         bld.recurse('systemd')
-    if not bld.env['IS_WINDOWS']:
+    if not bld.env['IS_WINDOWS'] and bld.env['HAVE_JACK_EXAMPLE_TOOLS']:
         bld.recurse('tests')
     if bld.env['BUILD_JACKDBUS']:
         bld.recurse('dbus')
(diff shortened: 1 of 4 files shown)


wscript
```diff
@@ -80,6 +80,7 @@ def options(opt):
     opt.add_option('--clients', default=256, type='int', dest='clients', help='Maximum number of JACK clients')
     opt.add_option('--ports-per-application', default=2048, type='int', dest='application_ports', help='Maximum number of ports per application')
     opt.add_option('--systemd-unit', action='store_true', default=False, help='Install systemd units.')
+    opt.add_option('--example-tools', type='choice', choices=['yes', 'no'], default='yes', help='Build and install jack-example-tools. Possible values: "yes" or "no"')
```
falkTX, line 83: why the custom choice instead of the usual `action='store_true', default=True,` like other arguments already use?

falkTX:
While this is greatly appreciated, the general style cleanup should not be present on the PR.
Arbitrarily changing the wscript files can break other existing and pending PRs, and generally makes it difficult to generate patches.
The removal of unused imports is fine with me though.

wscript
dvzrv: Because using a bool type option would not allow us what you communicated to me in regards to deprecation/removal in three (version) steps:
* add option to optionally build jack-example-tools (default to building the jack-example-tools)
* change option default to not build jack-example-tools
* remove option and files

A bool type option only allows to set the default (e.g. `False`) and then use it (unless there is some hidden magic in waf - would not surprise me - to automatically also have e.g. a `--not-example-tools` option or so) to set to `True` (using the action parameter).

dvzrv:
> While this is greatly appreciated, the general style cleanup should not be present on the PR. Arbitrarily changing the wscript files can break other existing and pending PRs, 

I understand. However, I have looked through the open PRs and most only add things to the file (with one exception, which hasn't been touched in two years and is also not rebased against develop).

>and generally makes it difficult to generate patches. The removal of unused imports is fine with me though.

Generating patches for simple additions on its own is also rather hard because there are hundreds of syntax problems in all the wscript files, which makes it hard to see the actual problems (e.g. IIO driver can not be built, even if we wanted to).

I have removed the commit, but will add it in a separate PR that fixes all the wscript files, as that allows also for better visibility during PRs. IMHO the existing PRs can and should be rebased on top of a change like that.

wscript
dvzrv: Okay, changed to use the autooptions stuff.

falkTX:
very nice, thanks for all the effort.

PR #826, opened 2022-01-07
Fix wscripts syntax

dvzrv:
This fixes the Python syntax in all wscripts (excluding man/wscript, as it is fixed by #825). The fixes include:

* all wscripts are allowed a line length of 120 chars
* unused imports are removed
* moves file-wide imports to top of file
* general syntax fixes according to flake8 recommendations
* adds a flake8 configuration file and runs flake8 against all wscripts in CI
* allows for the IIO driver to be built (picks up source files correctly)

This allows for better readability and detection of errors. Running flake8 can be included in CI as well.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 120
+output-file = flake8.txt
+exclude = .git,waflib,autooptions
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -308,7 +308,7 @@ jobs:
         apt-get source -d jackd2
         tar xf *.debian.tar.xz
         rm -rf debian/source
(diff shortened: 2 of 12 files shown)


falkTX:
Please rebase against latest develop branch, so we can test the windows builds that have now been fixed.
Thanks

falkTX:
The Ubuntu build fails because of a "VERSION=" parsing error that is now changed, I guess it just need to be replaced with "VERSION = ".

dvzrv:
> The Ubuntu build fails because of a "VERSION=" parsing error that is now changed, I guess it just need to be replaced with "VERSION = ".

That extraction is also very brittle (I also tried using sed on the file, but somehow this didn't work in CI...).
We could think about  moving the version information into a simple .env type file, where only `key=value` pairs are defined (without any quotes).
Aside from being less painful to parse (and set), as another side effect it would allow other build systems to make use of it in the future, while still only having to update one file to set the version strings :)

dvzrv:
> Please rebase against latest develop branch, so we can test the windows builds that have now been fixed. Thanks

Seems to work! \o/

dvzrv:
If CI passes, this is good to go!

PR #833, opened 2022-01-10
CI: Add FreeBSD builds through Cirrus CI.

0EVSG:
When the github application Cirrus CI (https://cirrus-ci.org/) is
installed, this build description file adds FreeBSD build checks to
changes and PRs in github.

There are four build tasks currently, a minimal build and one with all
options enabled on two different FreeBSD versions.

Unfortunately I don't see any other option of adding FreeBSD CI builds. But Cirrus CI has been quite unintrusive to me.

You can install the Cirrus CI Application from the Marketplace - choose the _public repositories only_ option (free) and select the repositories that use it. That will add the FreeBSD CI build checks to every change and PR that has this description file.

As a demo you can see what it looks like in my private repo:
https://github.com/0EVSG/jack2/tree/feature/freebsd_cirrus

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,29 @@
+task:
+  freebsd_instance:
+    matrix:
+      - image_family: freebsd-12-3
+      - image_family: freebsd-13-0
+
+  environment:
+    CFLAGS: -O2 -pipe -fPIC -fstack-protector-strong -fno-strict-aliasing -I/usr/local/include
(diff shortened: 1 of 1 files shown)


0EVSG:
If four build checks is too excessive, I can probably condense it into two or one.

falkTX:
Seems fine to me, ok to merge?

0EVSG:
I did another experiment with two builds packed into one check, but the results get a bit convoluted then. So I merge it as is and we'll adapt it if it gets annoying in some way.

@falkTX: If not already done, you'd have to install the Cirrus CI application - I don't have permission. Thanks!

PR #834, opened 2022-01-14
FreeBSD: Add missing OSS option --excl to man page.

0EVSG:
Did a short check of the man page and saw that --excl was missing. The other options seem to be fine, I left out the --device option because it is redundant and its short version -d may be confused with the driver backend -d.

While there, also improve the short description of the excl option in the OSS backend part of usage help and jack_control.

The options of the FreeBSD and Solaris OSS backends are kept compatible, intentionally.

Is there anything more to man page QA, apart from spellcheck?


falkTX:
> Is there anything more to man page QA, apart from spellcheck?

not sure if we even do that..

anyway, changes are minimal, let's merge

Issue #226, opened 2016-08-08
Jack error message


CreamyCookie:
I'm getting the following message when I use LibGDX with lwjgl3, which uses OpenAL-Soft, which interacts with libjack:

```
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
```

I get this whether or not I'm playing or initializing a sound. Audio works.

The following is [a comment from the OpenAL Soft developer](https://github.com/kcat/openal-soft/issues/53#issuecomment-237942514) about the issue:

> Those messages come from libjack, and unfortunately OpenAL Soft can't control them. OpenAL Soft tries to initialize the JACK backend first by trying to connect to a JACK server, in case that's the user's desired output method. If the server is not running, libjack will print that and fail (letting OpenAL Soft try PulseAudio next, then ALSA, etc), but **there doesn't seem to be a way to tell libjack that a failed connection isn't a problem and don't print that**.

I'm on Ubuntu 16.04. Feel free to ask me, if there's anything you need to know to fix this issue.

mseeber:
This is not about a failed connection, it is about creating a JACK client when no JACK server is running.
The default for open-al seems to be to try to create a client but disable starting a JACK server automatically, which seems reasonable and results in this message.
If you do not want to use JACK as audio backend, the right thing would be to disable it or move it to the lowest priority with the options open-al provides.

CreamyCookie:
Could there be an option / flag to hide these messages when trying to create a JACK client when no JACK server is running? 

> If you do not want to use JACK as audio backend, the right thing would be to disable it or move it to the lowest priority with the options open-al provides.

TBH I don't know anything about Jack (or OpenAL). I think Jack was preinstalled on Ubuntu 16.04., and then I got this error message out of the blue when using LibGDX (a game engine).

mseeber:
> Could you add an option / flag for the OpenAL guys to hide these messages when trying to create a JACK client when no JACK server is running?

I understand that the message is annoying but I don't know if this is a good idea. The reason is, that when using JACK the way it is used in the defaults the fact that no JACK server is running, the failed attempt to create a client becomes an error from a users point of view, because the settings assume, that a JACK server is running. So it does make sense to have a message in that case.

[...]

PR #835, opened 2022-01-16
API: Export missing symbols to jack library.

0EVSG:
The function pointers `jack_error_callback` and `jack_info_callback` are
offered by the jack.h header as non-weak API members, but they were not
exported to the jack library, only to the jackserver library.

This broke the Jack backend of the OpenAl Soft library when switching
from Jack1 to Jack2.

Some archeological references: #226, kcat/openal-soft#138, kcat/openal-soft#139

Alternative solution: Mark `jack_error_callback` and `jack_info_callback` as deprecated and export them as weak symbols. OpenAl Soft library already handles `jack_error_callback` like a weak symbol, and I'm not aware of any other software that is affected.

@falkTX: Do we need some additional reviewers / opinions on that?

### DIFF
--- a/common/JackError.cpp
+++ b/common/JackError.cpp
@@ -133,5 +133,5 @@ SERVER_EXPORT void silent_jack_error_callback(const char *desc)
 SERVER_EXPORT void silent_jack_info_callback(const char *desc)
 {}
 
-SERVER_EXPORT void (*jack_error_callback)(const char *desc) = &default_jack_error_callback;
-SERVER_EXPORT void (*jack_info_callback)(const char *desc) = &default_jack_info_callback;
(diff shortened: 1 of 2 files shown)


falkTX:
If this was working fine on jack1 and broke in jack2, then jack2 is broken.
So I am okay with just merging as-is. The PR fixes your issue, correct?

0EVSG:
This fixes the issue, I have it as a local patch here on FreeBSD. And it will be part of the official FreeBSD ports tree, when we update to 1.9.20. Suppose I'll just merge it then, we have plenty of time to test it before next release.

falkTX:
Yeah, now 3 months for next release.

PR #836, opened 2022-01-16
Fix build of jack-example-tools man pages

tartina:
Commit e8c0be1ce932bfb99fdd1a50e08fab187efed6ca changed the name
of the configuration environment to build the jack-example-tools,
but forgot to change it for man pages.


falkTX:
Thanks for the quick fix.

PR #844, opened 2022-01-30
Fixes for selected problems found by llvm scan-build.

0EVSG:
These problems were, amongst others, indicated by llvm scan-build (static analysis):

- Bad deallocation in JackMidiAsyncQueue. Should use delete[] for heap arrays.
- Bad semaphore allocation in midi_latency_test. Allocated pointer size instead of semaphore struct.
  - This should probably also go into jack-example-tools.
- Missing check for graph manager in JackAPI. Check GetGraphManager() for NULL there too.

Don't know if any of these affect real-life usage, but it's a good idea to fix them anyway.

### DIFF
--- a/common/JackAPI.cpp
+++ b/common/JackAPI.cpp
@@ -510,7 +510,7 @@ LIB_EXPORT int jack_port_tie(jack_port_t* src, jack_port_t* dst)
         jack_error("jack_port_tie called with ports not belonging to the same client");
         return -1;
     } else {
-        return manager->GetPort(mydst)->Tie(mysrc);
+        return (manager ? manager->GetPort(mydst)->Tie(mysrc) : -1);
     }
 }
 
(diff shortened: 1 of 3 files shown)


common/JackAPI.cpp
```diff
@@ -510,7 +510,7 @@ LIB_EXPORT int jack_port_tie(jack_port_t* src, jack_port_t* dst)
         jack_error("jack_port_tie called with ports not belonging to the same client");
         return -1;
     } else {
-        return manager->GetPort(mydst)->Tie(mysrc);
+        return (manager ? manager->GetPort(mydst)->Tie(mysrc) : -1);
```
falkTX, line 513: I think the check should be on the `if` above.
this should do it right?
```
if (manager || manager->GetPort(mysrc)->GetRefNum() != manager->GetPort(mydst)->GetRefNum()) {
```

falkTX: Meant to say `manager == nullptr` there

0EVSG: Well, it will prevent a possible crash, but it'll produce a misleading error message.
The other functions in the code use the same kind of in-place check, which is why I used that here.

falkTX: right, I see that now. a bit weird because it returns a generic -1 error without any info or warning log, but ok let's keep the code consistent.
will merge, thanks!

Issue #839, opened 2022-01-23
Build breaks on powerpc64le


pkubaj:
## Describe the bug

Build breaks on powerpc64le due to https://github.com/jackaudio/jack2/commit/dff7fa4fa8d25a8d3f21f4445a6496fd263e4a52

## Environment

* **JACK Version:** 1.9.20
* **Operating System:** FreeBSD 13.0-RELEASE
* **Installation:** ports tree

## Steps To Reproduce

Try to build on powerpc64le.

### Expected vs. actual behavior
```
[  3/181] Compiling common/JackDebugClient.cpp
[...]

PR #845, opened 2022-02-03
Fix alignas() on non-packed architectures (#839).

0EVSG:
Unbreak the build with Clang on architectures where JACK uses non-packed
data structures, like arm64 or powerpc. The alignment errors are exposed
there because:
 * The non-packed data structures require 8 byte alignment.
 * alignas() is not allowed to weaken alignment requirements to 4 bytes.
 * Clang enforces this according to standard, GCC ignores it.

Use an additional alignas() specifier to let the compiler choose between
minimum alignment (packed) and "natural" alignment (non-packed). This is
both standard c++11 and compatible with older GCC, which does not
propagate alignment requirements of packed substructures properly.

To the best of my knowledge, this does not change ABI on any platform.

Please note that JACK explicitly uses packed data structures on MacOS,
that's why it wasn't affected by this problem.

### DIFF
--- a/common/JackAtomicArrayState.h
+++ b/common/JackAtomicArrayState.h
@@ -122,7 +122,7 @@ class JackAtomicArrayState
         // fState[2] ==> request
 
         T fState[3];
-        alignas(UInt32) volatile AtomicArrayCounter fCounter;
+        alignas(UInt32) alignas(AtomicArrayCounter) volatile AtomicArrayCounter fCounter;
 
         UInt32 WriteNextStateStartAux(int state, bool* result)
         {
--- a/common/JackAtomicState.h
+++ b/common/JackAtomicState.h
@@ -94,7 +94,7 @@ class JackAtomicState
     protected:
 
         T fState[2];
-        alignas(UInt32) volatile AtomicCounter fCounter;
+        alignas(UInt32) alignas(AtomicCounter) volatile AtomicCounter fCounter;
         SInt32 fCallWriteCounter;
 
         UInt32 WriteNextStateStartAux()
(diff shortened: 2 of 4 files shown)


0EVSG:
@cme: I don't expect any changes to ABI from this PR, and the alignment assertions are left untouched.
But you may still be interested in testing this with shared x86 / arm64 memory on MacOS?

falkTX:
Small correction on the last note. x86 is not relevant for macOS anymore, the packed data is now for arm64 and x64 clients working with the same data structures. on macOS arm64 (maybe on other arm platforms too) there is a SIGBUS crash if accessing non-aligned shared memory data

0EVSG:
@falkTX: Understood. I meant the 64bit variant of x86 anyway, and I also think the original case for `alignas()` in #761 was about that?

Do you see any potential problems with this approach? I did some experiments with GCC 10 and GCC 6, and they just ignore the weaker alignment requirements in our cases. Which means this change doesn't affect the alignments or offsets. Although GCC is quite [buggy](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58601) regarding `alignas()`, nobody fixes that - so I expect it to be more or less consistently buggy.

falkTX:
I dont understand the GCC side well enough to say this is a problem or not.
We just need to test things, see if they work, and be happy until they dont :)

0EVSG:
Ok, would it be wise to invite someone else for testing / review?
Otherwise, if I don't hear anything negative from cme I'll just merge. And be happy :)

cme:
This looks good to me! I will try and test in the next few days, but it should work.

falkTX:
I have setups here that I can test the x86+x64 linux and x64+arm64 macOS, so let me verify this first.
ping me next week if I dont say anything

falkTX:
Tested to work on macOS universal builds. I had a similar issue with other software that was solved the same way, so I think this should be pretty safe.
Merging, thanks!

cme:
Awesome!

PR #846, opened 2022-02-06
fix JackWeakAPI on Windows

Be-ing:


common/JackWeakAPI.c
```diff
@@ -77,8 +82,11 @@ void *load_jack_function(const char *fn_name)
 {
     void *fn = 0;
     if (!libjack_handle) {
-        fprintf (stderr, "libjack not found, so do not try to load  %s ffs  !\n", fn_name);
-        return 0;
+        tryload_libjack();
+        if (!libjack_handle) {
+            fprintf (stderr, "libjack not found, so do not try to load  %s ffs  !\n", fn_name);
+            return 0;
+       }
```
Be-ing, line 89: Alternatively this could be done once when calling `jack_client_open` like weakjack does, but this was easier to implement. The application shouldn't call any JACK functions if that fails anyway.

PR #848, opened 2022-02-11
Update midiseq.c

strongheart1:
A typo in the usage string.
Alternatively, one can adapt the line 46.


dvzrv:
@strongheart1 thanks! It might be better to provide this as a PR to [jack-example-tools](https://github.com/jackaudio/jack-example-tools) instead, as the example-clients and tools will be removed from jack2 in two versions.

strongheart1:
Moved to https://github.com/jackaudio/jack-example-tools/pull/66

PR #852, opened 2022-02-17
pkgconfig: Use predefined variables in Libs and Cflags

Biswa96:
This fixes path relocation in mingw environment. Without this,
libdir and includedir are not shown in pkgconfig output. e.g.
* Without predefined variables:
  - pkgconf -cflags jack: No output
  - pkgconf -libs jack: -ljack64
* With predefined variables:
  - pkgconf -cflags jack: -IC:/msys64/mingw64/include
  - pkgconf -libs jack: -LC:/msys64/mingw64/lib -ljack64

Also official documentation suggests to use variables
https://people.freedesktop.org/~dbn/pkg-config-guide.html

### DIFF
--- a/jack.pc.in
+++ b/jack.pc.in
@@ -8,5 +8,5 @@ jack_implementation=jack2
 Name: jack
 Description: the Jack Audio Connection Kit: a low-latency synchronous callback-based media server
 Version: @JACK_VERSION@
-Libs: -L@LIBDIR@ -l@CLIENTLIB@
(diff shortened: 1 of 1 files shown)


falkTX:
That makes sense, thanks!
Please do the same for jackserver and jacknet pkg-config files too, thanks!

Biswa96:
But there is only jack.pc.in. Where can I find the other ones?

PR #857, opened 2022-03-29
JackFifo: Apply given init value in allocate

aditpape:
When using as sync opbject the given 'value' is meant to
already 'unblock' for that amount of Wait calls.

Signed-off-by: Andreas Pape <apape@de.adit-jv.com>


falkTX:
That makes sense, thanks!

PR #860, opened 2022-04-01
Fix crash in midi_latency_test

Teemperor:
`s` points to a constant string literal in the current code (which is in practice
just a pointer in some read-only memory). The next line `s[2] = c;` either
crashes or just gets completely removed by the compiler as it attempts to
write to read-only memory.


Teemperor:
Thanks for the quick review :) Opened a second PR here: https://github.com/jackaudio/jack-example-tools/pull/68

PR #861, opened 2022-04-01
Remove unused JackArgParser::GetArgv

Teemperor:
This is not called anywhere. Also this function doesn't do anything useful besides
leaking memory in its current form. I think the idea was to give this a
`char ***` (instead of `char **`) paramter that will be filled and then can
be used by the caller. But in its current form it just allocates a pointer
with the address being only used in the local variable (which then goes out
of scope and the memory is lost).

### DIFF
--- a/common/JackArgParser.cpp
+++ b/common/JackArgParser.cpp
@@ -121,26 +121,6 @@ namespace Jack {
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


Teemperor:
Note: The removed line `fill_n ( argv[i], fArgv[i].length() + 1, 0 );` is also a buffer overflow.

PR #863, opened 2022-04-13
Run flake8 on all wscript files and on jack_control

dvzrv:
.github/workflows/lint.yml:
Ensure that flake8 in CI is run on all wscript files and on
jack_control.


falkTX:
Thanks!

PR #876, opened 2022-06-06
IIO Driver requires the std namespace

flatmax:
Add the standard namespace to JackIIODriver.cpp otherwise build fails


imaami:
`using namespace std` is a blunt instrument and I'd personally steer clear of using it. Why not just prepend `std::` to where it might be needed instead of undoing the whole purpose of namespaces at the file scope?

Issue #869, opened 2022-05-17
jack_client_close makes jackd session unstable on macOS 12.3.1


steffendasenbrock:
I observed a similar bug as already reported [here](https://github.com/jackaudio/jack2/issues/841).

## Environment

* **JACK Version:** 1.9.21
* **Operating System:** macOS Monterey 12.3.1 (on MacBook Pro (14-inch,2021, M1 Pro))
* **Installation:** Installer from jackaudio.org

## Steps To Reproduce

I followed the instructions mentioned here [here](https://github.com/gisogrimm/test_jack_client_close)

### Expected vs. actual behavior

On macOS 12.3.1 I get:
```
% ./test_jack_client_close
0.037297 jack_client_open succeeded
0.037317 jack_set_process_callback(jc1,...) returned 0
0.047708 jack_activate(jc1) returned 0
1.074949 jack_client_open succeeded
1.074968 jack_set_process_callback(jc2,...) returned 0
1.094260 jack_activate(jc2) returned 0
2.098897 port type is "32 bit float mono audio"
JackMachSemaphore::TimedWait name = js502.cli1 usec = 9223372036854775807 err = (os/kern) aborted
SuspendRefNum error
JackClient::Execute error name = cli1
3.118966 jack_deactivate(jc1) returned 0
3.118980 jc1_process was called 143 times
3.118984 jc2_process was called 95 times
4.143361 jack_client_close(jc1) returned 0
5.147283 port type jc2 is "32 bit float mono audio"
JackMachSemaphore::TimedWait name = js502.cli2 usec = 9223372036854775807 err = (os/kern) aborted
SuspendRefNum error
JackClient::Execute error name = cli2
6.169666 jack_deactivate(jc2) returned 0
6.169711 jc1_process was called 143 times
6.169716 jc2_process was called 237 times
7.208441 jack_client_close(jc2) returned 0
8.252999 jack_client_open succeeded
9.258067 port type is "32 bit float mono audio"
10.279375 jack_client_close returned 0
10.279397 jc1_process was called 143 times
10.279399 jc2_process was called 237 times
```
 Expected output (Ubuntu 20.04.2, jack version: 1.9.12)
```
./test_jack_client_close
0.039246 jack_client_open succeeded
0.039310 jack_set_process_callback(jc1,...) returned 0
0.044882 jack_activate(jc1) returned 0
1.069785 jack_client_open succeeded
1.069831 jack_set_process_callback(jc2,...) returned 0
[...]

PR #884, opened 2022-07-10
Remove usage of 'U' mode bit for opening files in python

DaanDeMeyer:
The 'U' mode bit is removed in python 3.11. It has been
deprecated for a long time. The 'U' mode bit has no effect
so this change doesn't change any behavior.

See https://docs.python.org/3.11/whatsnew/3.11.html#changes-in-the-python-api

### DIFF
--- a/waflib/ConfigSet.py
+++ b/waflib/ConfigSet.py
@@ -312,7 +312,7 @@ class ConfigSet(object):
 		:type filename: string
 		"""
 		tbl = self.table
-		code = Utils.readf(filename, m='rU')
+		code = Utils.readf(filename, m='r')
 		for m in re_imp.finditer(code):
 			g = m.group
 			tbl[g(2)] = eval(g(3))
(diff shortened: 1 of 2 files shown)


nedko:
This issue was addressed by waf upstream 3 years ago: https://gitlab.com/ita1024/waf/-/commit/68997828c850ce7fb30b73b4adfde35053e539d1

Upgrading waf to latest should fix the py3.11 incompatibility.

cclauss:
Blocking https://github.com/Homebrew/homebrew-core/pull/119011

falkTX:
We do indeed need to update waf, but needs to be done carefully as there are custom changes in there. And I think extra modules too.
For now manually backporting fixes is quite welcome.

So thanks, merging now!

cclauss:
Can we please get a [release of jack2](https://github.com/jackaudio/jack2/releases) because this is one of the installs that is [blocking Homebrew's](https://github.com/Homebrew/homebrew-core/pull/119011) upgrade to Python 3.11 as the default Python?

falkTX:
If we manage to fix #869 then it would make sense to get a release out. Any help for that one is greatly appreciated!

Issue #388, opened 2018-10-28
Ringbuffers are not thread safe


nicholashaydensmith:
The ringbuffer code is not thread safe because variables `write_ptr` and `read_ptr` are read from and written to by different threads.  Not only are the `size_t` variables themselves subject to data hazard, the compiler and processor are free to reorder any of the operations inside of `jack_ringbuffer_read` and `jack_ringbuffer_write` because there is zero fencing. Consider the following scenario:

Producer thread: calls `jack_ringbuffer_write` and updates `write_ptr` before we actually memcpy any data and gets context switched out.
Consumer thread: calls `jack_ringbuffer_read` and thinks there is data to be consumed when in fact the producer thread hasn't actually written anything to memory.

The `volatile` keyword does not make anything threadsafe and shouldn't be used in this context.  `volatile` is really only useful for memory mapped IO and is never appropriate to use as a means for inter-thread communication.

Instead of using `volatile`, the `write_ptr` and `read_ptr` variables should be made atomic and use proper memory order semantics to ensure thread safety, https://en.cppreference.com/w/cpp/atomic/memory_order.

99.9% of the time I bet the code, as it stands, works perfectly fine so I understand this is probably not a super high priority issue.  Just thought I'd raise this issue for correctness purposes.  I would submit a pull request, but it would violate the terms and agreement with my current employer.

kmatheussen:
Hopefully it's only in theory that it would misbehave, at least on intel CPUs. But it's undoubtedly a real problem is if you use tsan, so this must be fixed.

Here's a disussion about it from a couple of years ago: http://linux-audio.4202.n7.nabble.com/Realtime-inter-thread-communication-td99157.html

I did supply a patch for the ringbuffer in that thread, but I've lost access to the account where it was uploaded. Anyway, it's an easy fix.

kmatheussen:
> The volatile keyword does not make anything threadsafe and shouldn't be used in this context. volatile is really only useful for memory mapped IO and is never appropriate to use as a means for inter-thread communication.

Actually, the microsoft c compiler guarantee instructions on volatile variables not to be reordered: https://preshing.com/20141024/my-multicore-talk-at-cppcon-2014/
https://msdn.microsoft.com/en-us/library/12a04hfd.aspx
Seems like Microsoft did the right thing there.

jcelerier:
> Before trying to fix this, we need to have a test case that proves the bad behavior.

[...]

PR #886, opened 2022-07-21
Fix ringbuffer thread safety on ARM. Fix #715 #388

### LINKED ISSUE #715: Thread safety of jack ringbuffer on ARM
## Describe the bug

This is a follow up to #388.

I am able to consistently reproduce an overbound read problem on a Raspberry Pi 4 using the test program in https://github.com/jackaudio/jack2/issues/388#issuecomment-453283106. Here is a sample output

```
...
---------------- 418756/1000000 ----------------
---------------- 418757/1000000 ----------------
message (0) != expected (82)
```

It happens almost deterministically for 1000000 iterations, but the exact timing is random. This program always terminates correctly on x86 and it only occurs on ARM.

The basic problem is that the write pointer is advanced before data is written completely. This is probably due to ARM's relaxed memory ordering. I am not sure about Apple Silicon, but it might also suffer from the same issue. A memory barrier here is not optional.

I will reproduce a cleaned up test program here:

```c
//          Copyright Jean Pierre Cimalando 2018.
// Distributed under the Boost Software License, Version 1.0.
//    (See accompanying file LICENSE or copy at
//          http://www.boost.org/LICENSE_1_0.txt)

// g++ -O2 -o test_case test_case.cc -ljack -lpthread

#include <jack/ringbuffer.h>
#include <memory>
#include <thread>
#include <stdio.h>
#include <stdlib.h>
#include <sched.h>

jack_ringbuffer_t *ring_buffer;
size_t capacity = 1024;
size_t message_count = 100;
size_t message_size = 100;

void thread1()
{
  jack_ringbuffer_t *rb = ::ring_buffer;
  const size_t message_size = ::message_size;
  const size_t message_count = ::message_count;
  std::unique_ptr<uint8_t[]> message_buffer(new uint8_t[message_size]);

  for (size_t i = 0; i < message_count;) {
    *(size_t *)message_buffer.get() = i;

    size_t can_write=jack_ringbuffer_write_space(rb);
    size_t count=0;
    if (can_write>=message_size) {
      count = jack_ringbuffer_write(rb, (const char *)message_buffer.get(), message_size);
    }
    if (count == message_size)
      ++i;
    else
      sched_yield();
  }
}

void thread2()
{
  jack_ringbuffer_t *rb = ::ring_buffer;
  const size_t message_size = ::message_size;
  const size_t message_count = ::message_count;
  std::unique_ptr<uint8_t[]> message_buffer(new uint8_t[message_size]);

  for (size_t i = 0; i < message_count;) {
    size_t can_read=jack_ringbuffer_read_space(rb);
    size_t count=0;
    if (can_read>=message_size) {
      count = jack_ringbuffer_read(rb, (char *)message_buffer.get(), message_size);
    }

    if (count == message_size) {
      size_t msg = *(size_t *)message_buffer.get();
      if (msg != i) {
        printf("message (%zu) != expected (%zu)\n", msg, i);
        exit(1);
      }
      ++i;
    }
  }
}

int main()
{
  size_t ntimes = 1000000;
  for (size_t i = 0; i < ntimes; ++i)
  {
    printf("---------------- %4zu/%4zu ----------------\n", i + 1, ntimes);
    jack_ringbuffer_t *rb = ::ring_buffer = jack_ringbuffer_create(::capacity);
    std::thread t1(thread1);
    std::thread t2(thread2);
    t1.join();
    t2.join();
    ::ring_buffer = nullptr;
    delete rb;
  }
  printf("-------------------------------------------\n");
  printf("success!\n");

  return 0;
}
```

## Environment

* **JACK Version:** 1.9.12 (`ringbuffer.c` hasn't been touched for 10 years)
* **Operating System:** Linux (Raspberry Pi OS)
* **Installation:** package manager

## Steps To Reproduce

```bash
$ g++ -O2 -o test_case test_case.cc -ljack -lpthread
$ ./test_case
```

Wait until it fails.

No overbound read should happen and the test program should always exit with a success on ARM.

Krasjet:
This patch addresses the thread safety problem of `jack_ringbuffer_t`
mentioned in #715 and #388. The overbound read bug caused by this problem
is impossible to reproduce on x86 due to its strong memory ordering, but
it is a problem on ARM and other weakly ordered architectures.

Basically, the main problem is that, on a weakly ordered architecture,
it is possible that the pointer increment after `memcpy` becomes visible
to the other thread before `memcpy` finishes:

	memcpy (&(rb->buf[rb->write_ptr]), src, n1);
	// vvv can be visible to reading thread before memcpy finishes
	rb->write_ptr = (rb->write_ptr + n1) & rb->size_mask;

If this happens, the other thread can read the remaining garbage values
in `rb->buf` due to be overwritten by the unfinished `memcpy`.

To fix this, an explicit pair of [release/acquire memory fences][1] is
used to ensure the copy on the other thread *happens after* the `memcpy`
finishes so no garbage values can be read.

[1]: https://preshing.com/20130922/acquire-and-release-fences/

### DIFF
--- a/common/jack/ringbuffer.h
+++ b/common/jack/ringbuffer.h
@@ -50,8 +50,8 @@ jack_ringbuffer_data_t ;
 
 typedef struct {
     char	*buf;
-    volatile size_t write_ptr;
-    volatile size_t read_ptr;
+    size_t	write_ptr;
+    size_t	read_ptr;
     size_t	size;
     size_t	size_mask;
     int	mlocked;
--- a/common/ringbuffer.c
+++ b/common/ringbuffer.c
@@ -27,6 +27,29 @@
 #endif /* USE_MLOCK */
 #include "JackCompilerDeps.h"
 
+/* Portable definitions for acquire and release fences */
+#if defined(_MSC_VER) && !defined(_M_AMD64) && !defined(_M_IX86) && !defined(_M_X64)
+	/* Acquire and release fences are only necessary for
+	 * non-x86 systems. In fact, gcc will generate no
+	 * instructions for acq/rel fences on x86. Hence, we only
+	 * use MemoryBarrier() for non-x86 systems with msvc */
+	#include <windows.h>
+	#define JACK_ACQ_FENCE() MemoryBarrier()
+	#define JACK_REL_FENCE() MemoryBarrier()
+#elif defined(__GNUC__)
+	#ifdef __ATOMIC_ACQUIRE
+		#define JACK_ACQ_FENCE() __atomic_thread_fence(__ATOMIC_ACQUIRE)
+		#define JACK_REL_FENCE() __atomic_thread_fence(__ATOMIC_RELEASE)
+	#else
+		/* Fallback to the legacy __sync builtin (full memory fence) */
+		#define JACK_ACQ_FENCE() __sync_synchronize()
+		#define JACK_REL_FENCE() __sync_synchronize()
+	#endif
+#else
+	#define JACK_ACQ_FENCE()
+	#define JACK_REL_FENCE()
+#endif
+
 typedef struct {
     char *buf;
     size_t len;
@@ -126,7 +149,7 @@ jack_ringbuffer_reset (jack_ringbuffer_t * rb)
 {
 	rb->read_ptr = 0;
 	rb->write_ptr = 0;
-    memset(rb->buf, 0, rb->size);
+	memset(rb->buf, 0, rb->size);
 }
 
 /* Reset the read and write pointers to zero. This is not thread
@@ -151,7 +174,7 @@ jack_ringbuffer_read_space (const jack_ringbuffer_t * rb)
 {
 	size_t w, r;
 
-	w = rb->write_ptr;
+	w = rb->write_ptr; JACK_ACQ_FENCE();
 	r = rb->read_ptr;
 
 	if (w > r) {
@@ -171,7 +194,7 @@ jack_ringbuffer_write_space (const jack_ringbuffer_t * rb)
 	size_t w, r;
 
 	w = rb->write_ptr;
-	r = rb->read_ptr;
+	r = rb->read_ptr; JACK_ACQ_FENCE();
 
 	if (w > r) {
 		return ((r - w + rb->size) & rb->size_mask) - 1;
@@ -199,6 +222,8 @@ jack_ringbuffer_read (jack_ringbuffer_t * rb, char *dest, size_t cnt)
 
 	to_read = cnt > free_cnt ? free_cnt : cnt;
 
+	/* note: relaxed load here, rb->read_ptr cannot be
+	 * modified from writing thread  */
 	cnt2 = rb->read_ptr + to_read;
 
 	if (cnt2 > rb->size) {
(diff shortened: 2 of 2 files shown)


falkTX:
Nice one.
Though since this is C code, we can simply change it to use C11 and use the new functions dedicated to this? Any reason to not go that route?

Krasjet:
The main problem is that atomics is an optional component in C11. Notably, MSVC [does not support it](https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/#whats-not), even though it is available as a C++ standard library. So to use the portable memory fence, we have to call a C++ function or convert jack_ringbuffer to C++, etc. Since the jack ringbuffer implementation is used elsewhere (pipewire, jack1, etc.), I think the best option is to keep it a self-contained C implementation.

Krasjet:
Do you think we should remove the `volatile` qualifier for `write_ptr` and `read_ptr` as well? The `__atomic_thread_fence` intrinsic should already prevent compiler reordering (I'll have to add a compiler barrier for x86 MSVC though).

0EVSG:
As I understand, the only case where `volatile` would matter in presence of `__atomic_thread_fence` is when being used from a signal handler (in the same thread). Not very likely, and the code doesn't seem robust in that sense anyway, but maybe we should keep them just in case.

FWIW, I did a short review of the changes and they look good to me, I'm not that experienced with memory barriers though. And I also checked that the thread fences are actually used on FreeBSD / Clang. So this is a go from me.

Krasjet:
There is actually a small overhead for using `volatile`. For example, you can see in the assembly generated [by gcc](https://godbolt.org/z/n8PT9fq1v) and [MSVC](https://godbolt.org/z/dseY54bs9) for `jack_ringbuffer_write` that every access to `write_ptr` issues a `mov` instruction when `volatile` is present
```asm
mov     ..., QWORD PTR [rbp+8]
```
but the write pointer is never going to be updated by the reading thread and it can be cached if we remove `volatile`.

The [`volatile` keyword in MSVC](https://docs.microsoft.com/en-us/cpp/cpp/volatile-cpp?view=msvc-170#end-of-iso-conformant) also has some non-portable semantics that enforces the release-acquire ordering when `/volatile:ms` is present. It's only enabled by default on x86, not ARM, but when the flag is present when compiling for ARM, the compiler also generates unnecessary `ldar` and `stlr` instructions that serves the same purpose as memory fences.

Krasjet:
I just checked that removing the `volatile` keyword in C doesn't seem to break the ABI, so I'll include the change as well.

An additional optimization for read/write space calculation is also added based on the discussion for zix ringbuffer [here](https://github.com/drobilla/zix/pull/1#issuecomment-1212687196).

Krasjet:
The changes should be ideally cherry-picked to jack1 as well. When it's ready.

elan:
I assume this will be merged at some point? 😅

falkTX:
yes of course. sorry for slow / little activity, lots of stuff happening at the same time lately.
there is a new incompatibility with new macOS too, still to be fixed.
these 2 things warrant a new release, maybe in december.

Krasjet:
Apparently the CI is down again for FreeBSD. Not a problem with the code.

0EVSG:
FreeBSD CI should be fixed now on the develop branch.

PR #889, opened 2022-08-15
JackClient: Lock before killing client notification thread

DeviLaxmii:
The client notification thread locks a mutex before reading/writing
from/to the socket. If the thread is killed while the lock is taken it
would leave the mutex dangling.

To avoid this take the mutex and then kill the client notification thread.

Signed-off-by: Laxmi Devi <Laxmi.Devi@in.bosch.com>


DeviLaxmii:
Hello,

I was able to reproduce the issue mentioned in : https://github.com/jackaudio/jack2/issues/395
With the below script
```
#!/bin/sh
export JACK_PROMISCUOUS_SERVER=jack; ulimit -r80 -l100000
a=0

while [ $a -lt 60 ]
do
   echo $a
   a=`expr $a + 1`
   aplay -v -d3  -Dentertainment_main /opt/platform/unit_tests/BigBuckBunny.wav
done
```

Kindly consider the patch which resolves the issue.

ReinholdH:
Thanks for the commit.
After searching for hours and almost next to give up, I found bug #395 and this commit #889.
Please add this commit to the next Jack update 1.9.22. 
In particular on Windows it is important when opening/closing Jack clients frequently.
Thanks.

ReinholdH:
This commit did not make it to Jack 1.9.22 despite it looks really simple.
Has there been any reason why?
Is this fix not complete?
To me this issue is a show stopper to effectively use Jack on Windows.
Thanks

falkTX:
thread sync always looks simple but never really is.
I need some time to do some testing of my own, for whch I dont have at the moment.
there were blockers on macOS and in waf that prevented jack2 from even being built at all, that took priority.
will see about other stuff like this once later

ReinholdH:
Thanks for your immediate response. I fully understand. Good luck for fixing such stuff.

ReinholdH:
Just to confirm, I have checked the issue again with 1.9.22. The locking still happens with Jack 1.9.22.

PR #932, opened 2023-05-04
Reproducibility  fixes

sten0:
Hello,

I'm a member of the Debian Multimedia Team, and we've recently applied these reproducibility fixes.  Thanks to @vagrantc

I hope that you'll consider applying them to your tree as well.

Regards,
Nicholas

### DIFF
--- a/man/fill_template
+++ b/man/fill_template
@@ -1,6 +1,6 @@
 #!/bin/sh
 d=""
-[ -z "$SOURCE_DATE_EPOCH" ] || d="--utc --date=@$SOURCE_DATE_EPOCH"
+[ -z "$SOURCE_DATE_EPOCH" ] || d="-u --date=@$SOURCE_DATE_EPOCH"
 
 if [ "$2" = "True" ]; then
   for i in *.0 ; do
--- a/man/fill_template
+++ b/man/fill_template
@@ -1,6 +1,6 @@
 #!/bin/sh
 d=""
-[ -z "$SOURCE_DATE_EPOCH" ] || d=--date=@$SOURCE_DATE_EPOCH
+[ -z "$SOURCE_DATE_EPOCH" ] || d="--utc --date=@$SOURCE_DATE_EPOCH"
 
 if [ "$2" == "True" ]; then
   for i in *.0 ; do
--- a/man/fill_template
+++ b/man/fill_template
@@ -2,7 +2,7 @@
 d=""
 [ -z "$SOURCE_DATE_EPOCH" ] || d="--utc --date=@$SOURCE_DATE_EPOCH"
 
-if [ "$2" == "True" ]; then
+if [ "$2" = "True" ]; then
   for i in *.0 ; do
     sed -e "s/!VERSION!/${1}/g" -e "s/!DATE!/`date $d '+%Y-%m-%d'`/g" < ${i} > ${i%%0}1
   done
--- a/man/fill_template
+++ b/man/fill_template
@@ -4,8 +4,8 @@ d=""
 
 if [ "$2" == "True" ]; then
   for i in *.0 ; do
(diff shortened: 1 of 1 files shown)


falkTX:
Please rebase against latest develop branch, github deprecated and removed macos-10.15 builders that we were still using, so macOS builds were not tested.

Changes look good to me, if it doesn't break mac builds then it should be good for merging.

sten0:
Filipe Coelho ***@***.***> writes:

> Please rebase against latest develop branch,

Done.

> github deprecated and removed macos-10.15 builders that we were still using, so macOS builds were not tested.
>
> Changes look good to me, if it doesn't break mac builds then it should be good for merging.
>

Does MacOS 10.15 use BSD `date` or GNU Coreutils' `date`?  Based on
these two resources, it looks like they may have switched to Coreutils:

https://www.unix.com/man-page/osx/1/date/
http://www.polarhome.com/service/man/?qf=DATE&tf=2&of=MacOSX&sf=1

If it's still BSD date, then I'll need to sacrifice a bit of readability
for compatibility.  On that note, do you prefer best practises from a
GNU perspective, or maximum BSD compatibility by default?  MacOS aside,
the changes proposed in this PR are slightly not BSD-ready.

Thanks!

falkTX:
jack2 is supported on macOS and FreeBSD, so it needs to work on those.
but in https://github.com/jackaudio/jack2/blob/develop/wscript#L835 we can see the manpages are only handled in Linux and FreeBSD.

@0EVSG do you have `date --utc` in FreeBSD? on macOS there is no such option, but mac tends to have old tools and maybe FreeBSD has that now?

0EVSG:
There's `date -u` but no `date --utc` - it may be possible to install gnu utils from ports, didn't check.

sten0:
0EVSG ***@***.***> writes:

> There's `date -u` but no `date --utc` - it may be possible to install gnu utils from ports, didn't check.

It's not best practises for GNU systems, but I'm guessing FreeBSD ports
would rather s/--utc/-u/ than install GNU Coreutils.

Thus I've pushed an additional commit to accommodate this case.  Maybe
it will make a future where MacOS has man pages for jack2 slightly more
likely to happen ;)

falkTX:
That seems fine, even on macOS `man date` shows a -u option:

```
-u   Display or set the date in UTC (Coordinated Universal) time
```

```
% date -u
Tue Jul  4 14:47:32 UTC 2023
```

@dvzrv as last check, since you handled the man page stuff before, do you see any potential issues with these changes?

0EVSG:
Thanks! `date -u` is even POSIX standard, just looked it up.
We're lucky to have so many standards and best practices to choose from :)

vagrantc:
I believe BSD style date also has a different invocation for using SOURCE_DATE_EPOCH.

https://reproducible-builds.org/docs/source-date-epoch/ has an example that should work with at least GNU date and BSD date:

    DATE_FMT="+%Y-%m-%d"
    SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
    BUILD_DATE=$(date -u -d "@$SOURCE_DATE_EPOCH" "$DATE_FMT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "$DATE_FMT" 2>/dev/null || date -u "$DATE_FMT")

0EVSG:
@vagrantc you're right, there's no `--date` argument either for FreeBSD `date`, but `-r` is supported.

If I read correctly, we only hit this code when `SOURCE_DATE_EPOCH` is set. Which is not the case at the moment. And I don't know of any significant efforts towards reproducible builds on any *BSD. No idea about macOS.

Still, it might be easier to do this in a portable fashion right now, than to deal with possible fallout in the future.

PR #948, opened 2023-07-24
Adit control api backports

amiartus:
Hello,

please consider accepting following linux specific back-ports developed by ADIT audio team. The patches allow proper responsiveness to some failure states on linux and fix the hang ups that would occur otherwise.

Best Regards,
Adam Miartus

### DIFF
--- a/common/JackControlAPI.cpp
+++ b/common/JackControlAPI.cpp
@@ -26,6 +26,11 @@
 #include <pthread.h>
 #endif
 
+#ifdef __linux__
+#include <poll.h>
+#include <sys/signalfd.h>
+#endif
+
 #include "types.h"
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


nedko:
These look good to me. Thank you! The added [signal setup](https://github.com/jackaudio/jack2/pull/948/commits/58ab7eea9a090ad1f2710566fd77b1f662475c21#diff-46b84821282e2ca7c36e5dc90e9b70acff9ca51dcc9db5e20bc045ba27705cbfR689) should really be moved to jackctl_setup_signals() in a followup commit (these ones are cherry picks and it is good to keep them like this instead of rewriting git history). For jackdbus at least, it would be good to have epoll() call on the side of the control API user/caller (jackd, jackdbus). Extending the Control API with notion of either libevent (should work on Windows too), or similar abstraction, will achieve this goal.

nedko:
I added jackctl_finish_signals() to the API: https://github.com/LADI/jack2/commit/ae7b5d0dda4205adeccab85e6b1ca99aab2300fc

It is a function to call for cleaning,
after waiting on a signal set,
to undo the jackctl_setup_signals() and restore to previous state.

amiartus:
thank you for reviewing and refactoring. looks better to me as it is now after your changes.

common/JackControlAPI.cpp
```diff
@@ -679,16 +714,70 @@ jackctl_setup_signals(
...
+    }
+    eventFD = pfd[JackEventFD].fd;
+    pfd[JackEventFD].events = POLLIN;
+
+#endif
 
     while (waiting) {
     #if defined(sun) && !defined(__sun__) // SUN compiler only, to check
         sigwait(&sigmask->signals);
+        fprintf(stderr, "Jack main caught signal\n");
```
imaami, line 747: Should a library API function such as this really print errors directly to `stderr` like this? I'm not asking this rhetorically, I genuinely don't know what the expectation is regarding jackctl API behavior. I wanted to bring this up just in case it matters.

common/JackControlAPI.cpp
```diff
@@ -679,16 +714,70 @@ jackctl_setup_signals(
...
+    #elif defined(__linux__)
+        err = poll(pfd, JackFDCount, -1);
+        if (err < 0) {
+            if (errno == EINTR) {
+                continue;
+            } else {
+                fprintf(stderr, "Jack : poll() failed with errno %d\n", -errno);
```
imaami, line 754: Regardless of how error reporting should work here, printing the negation of `errno` seems bizarre. AFAIK only kernel code uses negative `errno` values due to implementation reasons, but in this situation I can't see how it's anything but a magic number that also happens to be even less useful than an unmodified `errno`. If there must be `stderr` output, why not just call `strerror(errno)`?

common/JackControlAPI.cpp
```diff
@@ -712,6 +801,16 @@ jackctl_wait_signals(jackctl_sigmask_t * sigmask)
         // bugs that cause segfaults etc. during shutdown.
         sigprocmask(SIG_UNBLOCK, &sigmask->signals, 0);
     }
+
+#ifdef __linux__
+fail:
+    for(int i = 0; i < JackFDCount; i++) {
```
imaami, line 807: This and a handful of other lines don't conform to the style of the rest of the code. AFAIK there should be a space after `if`, `for`, etc.

common/JackControlAPI.cpp
```diff
@@ -679,16 +714,70 @@ jackctl_setup_signals(
...
 jackctl_wait_signals(jackctl_sigmask_t * sigmask)
 {
-    int sig;
+    int sig = 0;
     bool waiting = true;
+#ifdef __linux__
+    int err;
+    struct pollfd pfd[JackFDCount];
+    struct signalfd_siginfo si;
+    memset(pfd, 0, sizeof(pfd));
```
imaami, line 723: Each `struct pollfd::fd` should be initialized to -1, not 0.

common/JackControlAPI.cpp
```diff
@@ -712,6 +801,16 @@ jackctl_wait_signals(jackctl_sigmask_t * sigmask)
         // bugs that cause segfaults etc. during shutdown.
         sigprocmask(SIG_UNBLOCK, &sigmask->signals, 0);
     }
+
+#ifdef __linux__
+fail:
+    for(int i = 0; i < JackFDCount; i++) {
+        if(pfd[i].fd != 0) {
```
imaami, line 808: This is wrong. If `eventfd()` fails then `close(-1)` gets called here. Initialize all fds to -1 and compare against that as the canonical "invalid fd" value.

imaami (changes requested):
Some observations. I don't really know JACK-specific details, please double-check my conclusions with that in mind.

common/JackControlAPI.cpp
amiartus: not sure either, but logging is not consistent all over jack2, some functions use printf some jack_log/jack_error. Its not clear to me when to use what...

amiartus:
thanks @imaami I can have a look at this over the holidays

PR #951, opened 2023-07-30
Add a couple of missing headers in macosx sources

barracuda156:
It was forgotten for this one.


barracuda156:
Converted to draft, because another header is missing in other source file. Will update soon.

P. S. `<algorithm>` is needed for `std::find`.

barracuda156:
@falkTX Could you please review? Regardless of ICE with GCC which hopefully gonna be fixed (and relevant only for 32-bit), these are needed.

PR #943, opened 2023-07-06
FreeBSD: Rewrite OSS backend, introduce sosso library

0EVSG:
This is a complete rewrite of the OSS backend for FreeBSD - none of this
affects other systems or touches code outside of the `freebsd` directory.
Main motivation was to make better use of my RME RayDAT PCIe sound
card and its low-latency capabilities. This endeavor quickly escalated into
separate testing programs and then a C++ header-only library to manage
OSS channels. The result of which is hereby used to implement the
FreeBSD OSS backend of JACK, bringing some substantial improvements
over the old code.

### The synced OSS operation (sosso) library

As a header-only C++ library, synced OSS operation (sosso) is used to
clean up and separate the low-level handling of FreeBSD OSS devices
in JACK. Features include:

 * Supports both read() / write() and mmap() IO operation.
 * Adaptive polling, better suited for low-latency requirements.
 * Internal double buffer to avoid troubles with OSS buffer resize.
 * Closely monitors progress and drift for each channel (+/- 1ms).
 * Facilitates drift correction through buffer offsets.
 * Depends on C++ standard library and system headers, nothing else.

Although the sosso library is somewhat tailored to the needs of JACK,
it was developed separately and will eventually be published and / or
used in other projects. Therefore the headers follow a different
formatting style and are more liberally licensed (ISC).

### Rewrite the OSS backend using the sosso library

Make use of the sosso library to handle the low-level operation of
OSS devices on FreeBSD. Further separate this part into the
JackOSSChannel class, which should make it easier to reuse the code
for JackOSSAdapter (not planned yet).
JackOSSChannel features an additional assist thread which helps to
process the OSS device when the driver thread is busy with the JACK
dependency graph.
Combined with the improvements of the sosso library, these changes
bring JACK into the realm of real-time audio processing on FreeBSD,
for select PCI sound cards (4-6ms round trip latency).

PR #952, opened 2023-07-30
JackPosixProcessSync: Fix mutex owner on TimedWait() timeout.

0EVSG:
Per POSIX definition, pthread_cond_timedwait() re-acquires the mutex when a timeout occurs and ETIMEDOUT is returned. In that case also mark the waiting thread as owner again, for consistency. Otherwise subsequent attempts to unlock the mutex will fail, leaving the mutex locked forever.

The return value of TimedWait() is still false on timeout, the same as for other platform's implementations.
But AFAICT, JackPosixProcessSync::TimedWait() is only used by my new FreeBSD OSS backend, see #943.

### DIFF
--- a/posix/JackPosixProcessSync.cpp
+++ b/posix/JackPosixProcessSync.cpp
@@ -122,7 +122,7 @@ bool JackPosixProcessSync::TimedWait(long usec)
     time.tv_nsec = (next_date_usec % 1000000) * 1000;
 
     res = pthread_cond_timedwait(&fCond, &fMutex, &time);
-    if (res != 0) {
+    if (res != 0 && res != ETIMEDOUT) {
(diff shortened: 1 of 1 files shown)


0EVSG:
Same here, should I just merge this if nobody's got time for a review?

falkTX:
Would be nice to confirm with some automated testing that we do not get regressions, but since as you say this is only used for BSD the breakge is more minimal and I can assume you tested it.
Feel free to merge yes

0EVSG:
It's difficult to test for regression when the code isn't even in use ;-)

I only discovered this problem when testing #943 - the timed wait treated a timeout like an error. Which left the mutex locked, but with ownership unset, and thus prevented any further unlocking of the mutex. I'm pretty sure that behavior was wrong and this patch at least improved it. Thread synchronization has been robust since then, so I'm gonna merge.

PR #953, opened 2023-08-29
Update to waf 2.0.26

nphilipp:
This makes waf compatible with Python 3.12 again.

### DIFF
--- a/waf
+++ b/waf
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python
 # encoding: latin-1
 # Thomas Nagy, 2005-2018
 #
@@ -32,7 +32,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 import os, sys, inspect
 
-VERSION="2.0.12"
+VERSION="2.0.26"
 REVISION="x"
 GIT="x"
 INSTALL="x"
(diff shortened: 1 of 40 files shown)


falkTX:
I was about to do this today, good timing!

There are some small changes to import though, regarding macOS versioning (need to keep it compatible with old releases)
Please import the waflib changes in 0f2e3b2e830e7b9ea03bb67a00c1eec4deab0855 and dc6c9959edb6c9807200aa7f1bbc894b59d74b64

we should probably combine these in a patch and keep it in the waflib dir, for the next time we need to update again...

nphilipp:
I’ll do that, but out of curiousity: isn’t it possible to integrate these changes upstream in waf so you don’t have to care about it any longer?

falkTX:
> I’ll do that, but out of curiousity: isn’t it possible to integrate these changes upstream in waf so you don’t have to care about it any longer?

I dont think that is doable, this is a very special case due to previous jack2 builds being created using xcode in a very specific way (which used a single machine with a very specific setup too, which has since died) and then we moved macOS builds to using waf.

There were talks and even some initial code for switching to meson, as waf relying on python makes the build system break very easily..
So I do not think it is worth to bother waf developers with this, even more for a situation that is likely temporary

falkTX:
Thanks for PR! it is merged now

PR #963, opened 2023-12-13
jackdbus: Stop recurrent wakeups when no save is pending

### LINKED ISSUE #962: Jackdbus wakes up every 200ms
Even when jackdbus is just listening for DBus events it continuously wakes up every 200 ms.

## Environment

* **JACK Version:** 1.9.22
* **Operating System:** Linux (openSUSE Tumbleweed 20231212)
* **Installation:** Distribution packages

## Steps To Reproduce

Start jackbus
Attach strace

When jackdbus is idle, it should not wake up.

Waking up, even when there are no actual events pending, shortens runtime on battery. This is especially important as jackdbus is often just running in the background.

https://github.com/jackaudio/jack2/blob/886b35c3bda019a2e83102c0711bbaa475efde7e/dbus/jackdbus.c#L950

The only actual need for the timed wakeup is a pending save.

StefanBruens:
Most processing happens as a result of an incoming DBus event. The only case when a timed processing
is needed is due to an pending save.

Instead of implementing a full event loop just run the timed loop while a save is pending, and use an
infinite timeout otherwise.

Fixes: #962

### DIFF
--- a/dbus/controller.c
+++ b/dbus/controller.c
@@ -783,7 +783,7 @@ jack_controller_destroy(
     free(controller_ptr);
 }
 
-void
+bool
 jack_controller_run(
     void * context)
 {
@@ -791,7 +791,7 @@ jack_controller_run(
 
     if (controller_ptr->pending_save == 0)
     {
-        return;
+        return false;
     }
 
     if ((ut = uptime()) < 0)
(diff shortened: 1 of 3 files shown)


falkTX:
The PR changes seem fine, but if one is using jack the idea of saving battery is non-sensical.
JACK will keep the audio device active at all times on a very quick / small buffer size, this will save very little

StefanBruens:
> The PR changes seem fine, but if one is using jack the idea of saving battery is non-sensical. JACK will keep the audio device active at all times on a very quick / small buffer size, this will save very little

Running jackdbus does not equal running jack. People will have jackdbus running in the background continously, even if they use jack just sometimes.

PR #967, opened 2024-01-19
Use XDG_STATE_HOME for jackdbus logs

### LINKED ISSUE #402: Make the log output directory XDG compliant or configurable
I noticed that the path to the log output for jack is hardcoded:
https://github.com/jackaudio/jack2/blob/cc8576a7ab62caab10c7ddd7309367cf797c5e69/dbus/jackdbus.c#L713

I know there is no `$XDG_LOG_HOME` in the standard so I suggest making it configurable through some custom environment variable like `$JACKLOG` or something and then falling back on `$XDG_DATA_HOME/jack` or `$XDG_RUNTIME_DIR`

spec:
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

keithbowes:
Fixes #402 

In that bug, it's mentioned that Debian has an $XDG_STATE_HOME extension for that purpose.  Well, it's been added to [version 0.8 of the specification](https://specifications.freedesktop.org/basedir-spec/0.8/) (8 May 2021) with the description:

> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> *    actions history (logs, history, recently used files, …)
> *    current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

It's been a part of the official specification for almost three years and other applications use it, so it should be safe to use it and not add unnecessary files and directories to $HOME.

### DIFF
--- a/dbus/jackdbus.c
+++ b/dbus/jackdbus.c
@@ -694,7 +694,7 @@ pathname_cat(const char *pathname_a, const char *pathname_b)
 bool
 paths_init()
 {
-	const char *home_dir, *xdg_config_home, *xdg_state_home;
+    const char *home_dir;
     
     home_dir = getenv("HOME");
     if (home_dir == NULL)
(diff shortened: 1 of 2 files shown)


dbus/jackdbus.c
```diff
@@ -694,7 +694,7 @@ pathname_cat(const char *pathname_a, const char *pathname_b)
 bool
 paths_init()
 {
-	const char *home_dir, *xdg_config_home, *xdg_log_home;
+	const char *home_dir, *xdg_config_home, *xdg_state_home;
```
imaami: Tabs are the wrong indentation style here. The whole file contains just a handful of tab-indented lines (which have been like that since 2008), all the rest is indented with 4 spaces. Now would be a good time to _not_ replicate the exact original indentation for the changed lines, but instead to fix all of it. Please use 4 spaces for your changes,, and also fix the remaining old tab-indented lines in the function, there's like 5 of them.

dbus/jackdbus.c
```diff
@@ -709,17 +709,21 @@ paths_init()
 	    if (!(xdg_config_home = pathname_cat(home_dir, DEFAULT_XDG_CONFIG))) goto fail;
 	}
 
-    if (!(xdg_log_home = pathname_cat(home_dir, DEFAULT_XDG_LOG))) goto fail;
+    xdg_state_home = getenv("XDG_STATE_HOME");
+    if (xdg_state_home == NULL)
+    {
+        if (!(xdg_state_home = pathname_cat(home_dir, DEFAULT_XDG_STATE))) goto fail;
```
imaami: This leaks memory and is problematic in general. A pointer from `getenv()` shouldn't be `free()` d, but `pathname_cat()` returns an allocated buffer pointer which definitely should. If you use a single pointer variable like this, you lose information about whether you need to call `free()` on it before returning.

That being said, none of the temporary allocations from `pathname_cat()` are ever freed. The whole thing seems to have been leaky since 2008.

PR #976, opened 2024-05-13
Update README

Dekker500:
Removed invalid link. Destination no longer contains any references to JACK2


falkTX:
It leads to 404 indeed, thanks!

PR #991, opened 2024-11-03
Change to JackTools::GetUID() on Windows that fixes metadata

kinggeek:
Change to JackTools::GetUID() on Windows that fixes metadata creating a new BDB on every API call.

JackTools::GetUID() would return the PID of the calling process. (I think this was a stub because there is no Windows equivalent). The linux version appears to return a linux UID. This patch does something similar on Windows so as to create/open the same DB on Windows and metadata now works correctly.

### DIFF
--- a/common/JackTools.cpp
+++ b/common/JackTools.cpp
@@ -69,8 +69,19 @@ namespace Jack {
     int JackTools::GetUID()
     {
 #ifdef WIN32
-        return  _getpid();
-        //#error "No getuid function available"
+        HANDLE tokenHandle = nullptr;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &tokenHandle)) {
+            return 0;
+        }
+
+        TOKEN_STATISTICS tokenStats;
+        DWORD returnLength;
+        if (!GetTokenInformation(tokenHandle, TokenStatistics, &tokenStats, sizeof(tokenStats), &returnLength)) {
+            CloseHandle(tokenHandle);
+            return 0;
+        }
+
+        return(tokenStats.AuthenticationId.LowPart);
 #else
         return geteuid();
 #endif


common/JackTools.cpp
```diff
@@ -69,8 +69,19 @@ namespace Jack {
...
 #ifdef WIN32
-        return  _getpid();
-        //#error "No getuid function available"
+        HANDLE tokenHandle = nullptr;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &tokenHandle)) {
+            return 0;
+        }
+
+        TOKEN_STATISTICS tokenStats;
+        DWORD returnLength;
```
imaami, line 78: Initialize `tokenStats` and `returnLength`:

```c++
        TOKEN_STATISTICS tokenStats{};
        DWORD returnLength = sizeof tokenStats;
```

common/JackTools.cpp
```diff
@@ -69,8 +69,19 @@ namespace Jack {
...
+        if (!GetTokenInformation(tokenHandle, TokenStatistics, &tokenStats, sizeof(tokenStats), &returnLength)) {
+            CloseHandle(tokenHandle);
```
imaami, line 80: You're only closing the handle if `GetTokenInformation()` succeeds, it should be closed every time.

common/JackTools.cpp
```diff
@@ -69,8 +69,19 @@ namespace Jack {
...
+            return 0;
+        }
+
+        return(tokenStats.AuthenticationId.LowPart);
```
imaami, line 84: Nitpick: `return` is not a function call, don't make it look like one.

imaami (changes requested):
A few things should be fixed. For completeness' sake here's my suggestion:

```c++
    int JackTools::GetUID()
    {
#ifdef WIN32
        int id = 0;
        HANDLE tokenHandle = nullptr;

        if (OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &tokenHandle)) {
            TOKEN_STATISTICS tokenStats{};
            DWORD len = sizeof tokenStats;

            if (GetTokenInformation(tokenHandle, TokenStatistics,
                                    &tokenStats, sizeof tokenStats, &len)) {
                id = static_cast<int>(tokenStats.AuthenticationId.LowPart);
            }

            CloseHandle(tokenHandle);
        }

        return id;
#else
        return geteuid();
#endif
    }
```

PR #992, opened 2024-11-03
Drop use of non-initialized variables

sthibaul:


imaami (approved):
Looks good to me. Btw, this same bug is also in `JackBoomerDriver::DisplayDeviceInfo()`.

imaami:
@falkTX this is about as trivial of a fix as it gets, would be nice to see it merged.

sthibaul:
> Btw, this same bug is also in JackBoomerDriver::DisplayDeviceInfo().

Indeed, now fixed there as well.

imaami (approved):
LGTM

falkTX:
I guess no one tested this before :shrug: 
Anyhow makes sense, the error printed by this is nonsense. Thanks!

PR #1003, opened 2025-05-30
Don't include headers within `extern "C"`

thesamesam:
Including C headers inside of extern "C" breaks use from C++. Hoist the includes of standard C headers above the block so we don't try to mangle the stdlib.

See https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/2377 in PipeWire where they bundle some JACK headers, so came across this.


falkTX:
Makes sense, thanks!

thesamesam:
Thanks!

